PR #1, opened 2020-04-09
Token fixes

arscan:
Fixes upload and adds example bearer token to README for new token format.


arscan:
I think I brought along a commit that wasn't brought up to `master` on GitHub, but it seems we need it, and it is a pretty benign update so I didn't fix.

PR #2, opened 2020-04-09
Fi 355 make db start populated

hershilpatel:
Updated the docker-compose to use Dockerfile.database
Dockerfile.database now copies an initdb.sql file so the data will load on first start up
Updated README to reflect changes and be more accurate
deleted unecessary files


arscan:
what is the command you use to generate `initdb.sql`?

arscan:
Looks good to me too -- I deleted the volume, restarted, and it worked.  @hershilpatel could you please update the readme with a section on how to regenerate initdb.sql based on the current snapshot of what is in the database?  Mainly I'm just looking for the sql command documented :).  Once that's done I'll approve and merge.  Thanks!

hershilpatel:
docker exec <DOCKER_CONTAINER_ID> pg_dump -U postgres postgres  > initdb.sql

hershilpatel:
Should I include this command somewhere?

okeefm:
You can do `docker-compose exec db` instead of looking up the container id @hershilpatel

hershilpatel:
The docker compose still uses the Dockerfile.database, we have a copy step to copy initdb.sql into the correct folder

PR #7, opened 2020-04-17
Make root url reference server

hershilpatel:
Changed the root url for the fhir reference server to <base>/reference-server , so now each servlet is under the url reference-server

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,8 +3,8 @@
 This is a FHIR reference implementation server supporting the US Core R4 IG and SMART Launches.
 
 By default, you can browse the server at
-[http://localhost:8080](http://localhost:8080), and the FHIR endpoint is at
-[http://localhost:8080/r4](http://localhost:8080/r4)
+[http://localhost:8080](http://localhost:8080/reference-server), and the FHIR endpoint is at
+[http://localhost:8080/r4](http://localhost:8080/reference-server/r4)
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -3,8 +3,8 @@
 This is a FHIR reference implementation server supporting the US Core R4 IG and SMART Launches.
 
 By default, you can browse the server at
-[http://localhost:8080](http://localhost:8080), and the FHIR endpoint is at
-[http://localhost:8080/r4](http://localhost:8080/r4)
+[http://localhost:8080](http://localhost:8080/reference-server), and the FHIR endpoint is at
+[http://localhost:8080/r4](http://localhost:8080/reference-server/r4)
```
arscan, line 7:
```suggestion
[http://localhost:8080/reference-server](http://localhost:8080/reference-server), and the FHIR endpoint is at
[http://localhost:8080/reference-server/r4](http://localhost:8080/reference-server/r4)
```

arscan:
The last commit here didn't seem to fix the port problem I was seeing, so perhaps it requires updates in a few places.

arscan:
I messed up my own test setup, which caused it to fail (didn't put in the IP when starting the launch, instead of localhost, even though i had configured inferno to expect an ip).  Tried again and it looks good.

PR #11, opened 2020-05-11
allow service calls with no resource

hershilpatel:
if you don't specify any resource, the fhir reference server will still allow it through

### DIFF
--- a/src/main/java/org/mitre/fhir/authorization/FakeOauth2AuthorizationInterceptorAdaptor.java
+++ b/src/main/java/org/mitre/fhir/authorization/FakeOauth2AuthorizationInterceptorAdaptor.java
@@ -53,7 +53,7 @@ public class FakeOauth2AuthorizationInterceptorAdaptor extends InterceptorAdapte
 
 
     String[] scopesArray = scopes.split(" ");
-    List<String> validResources = new ArrayList<String>();
+    List<String> grantedResources = new ArrayList<String>();
 
 
     for (String currentScope : scopesArray) {
@@ -67,9 +67,9 @@ public class FakeOauth2AuthorizationInterceptorAdaptor extends InterceptorAdapte
 
 
         if (scopeAfterSlashParts.length == 2) {
-          validResources.add(scopeAfterSlashParts[0]);
+          grantedResources.add(scopeAfterSlashParts[0]);
         } else {
-          validResources.add(scopeAfterSlash);
+          grantedResources.add(scopeAfterSlash);
         }
       }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/mitre/fhir/authorization/FakeOauth2AuthorizationInterceptorAdaptor.java
```diff
@@ -77,10 +77,13 @@ public boolean incomingRequestPostProcessed(RequestDetails requestDetails,
...
 
-    if (!validResources.contains("*")
-        && !validResources.contains(resource)
+    if (!validResources.contains("*") && !validResources.contains(resource)
         && !("Patient".equals(resource))) {
-      throw new InvalidScopesException(resource);
+      if (resource != null) {
+        throw new InvalidScopesException(resource);
+      } else {
+        Log.error("No Resource was provided");
```
arscan, line 85: This isn't an error, because it is doing what it is supposed to be doing.  But it is unsafe if you were to do this on a production system for HAPI because this implementation isn't inspecting resources returned in the paged bundle.  So, you could do a warning, but you are going to get a ton of warnings for a properly working system.

src/test/java/org/mitre/fhir/authorization/TestAuthorization.java
```diff
@@ -153,6 +153,7 @@ public void testInterceptor() {
     ourClient.create().resource(pt).execute();
   }
 
+  //TODO Readd this test when support for scope checking is readded
```
arscan, line 156: ?

Did you forget this comment before?  I thought you are checking scopes?

src/main/java/org/mitre/fhir/authorization/FakeOauth2AuthorizationInterceptorAdaptor.java
```diff
@@ -77,10 +77,13 @@ public boolean incomingRequestPostProcessed(RequestDetails requestDetails,
 
     String resource = requestDetails.getResourceName();
 
-    if (!validResources.contains("*")
-        && !validResources.contains(resource)
+    if (!validResources.contains("*") && !validResources.contains(resource)
```
arscan, line 80: I would rename `validResources` to `grantedResources` because 'valid' has a particular meaning in FHIR.  And change `resource` to `requestedResource`.  I'm not sure why 'Patient' is always given a pass, technically you could have a patient-level access but not be given access to the Patient resource (the Patient resource is 'demographics' about a patient -- name, location, contact... perhaps you want to keep it anonymous).

PR #12, opened 2020-05-19
added ability to choose which scopes you approve the user

hershilpatel:
now after the user selects a patient, it takes them to a page where they can select which of the scopes they would like to approve

note the ui is pretty simple, so if anyone thinks we should update it to make it loop better let me know!


arscan:
I created a ticket for considering UI updates -- at least adding a label tag around the text for each button so it is easier to click.  But for now works great!

PR #17, opened 2020-06-24
updated refresh token to contain scope instead of the token itself

hershilpatel:
Removed the encoded scopes from the bearer token, and instead store that information on the server.  This means that someone with the token cannot change the scopes by understanding how the token is structured and the token is simpler in general.  Current this change is only for the bearer token, and not the code (which is not stored on the reference server at the moment, but will be updated in https://oncprojectracking.healthit.gov/support/browse/FI-939

### DIFF
--- a/src/main/java/org/mitre/fhir/authorization/AuthorizationController.java
+++ b/src/main/java/org/mitre/fhir/authorization/AuthorizationController.java
@@ -134,7 +134,7 @@ public class AuthorizationController {
         patientId = new String(Base64.getDecoder().decode(encodedPatientId));
       }
 
-      if ((code != null && FhirReferenceServerUtils.SAMPLE_CODE.equals(actualCodeOrRefreshToken))) {
+      if (FhirReferenceServerUtils.SAMPLE_CODE.equals(actualCodeOrRefreshToken)) {
         return generateBearerTokenResponse(request, clientId, scopes, patientId);
       }
 
--- a/src/main/java/org/mitre/fhir/authorization/AuthorizationController.java
+++ b/src/main/java/org/mitre/fhir/authorization/AuthorizationController.java
@@ -117,8 +117,7 @@ public class AuthorizationController {
     if (code != null) {
       fullCodeString = code;
 
-
-      // the provided code is actualcode.scopes
+      // the provided code is in the format <ACTUAL_CODE>.<SCOPES>
       String[] fullCode = fullCodeString.split("\\.");
 
       String actualCodeOrRefreshToken = fullCode[0];
@@ -152,8 +151,6 @@ public class AuthorizationController {
             "Refresh Token " + refreshTokenValue + " was not found");
       }
     }
-
-
 
     throw new ResponseStatusException(HttpStatus.BAD_REQUEST, "Invalid code");
   }
@@ -214,7 +211,6 @@ public class AuthorizationController {
       throw new ResponseStatusException(HttpStatus.UNAUTHORIZED, "No patients found");
     }
 
-    // get their id
     List<String> scopesList = Arrays.asList(scopes.split(" "));
 
     if (scopesList.contains("launch") || scopesList.contains("launch/patient")) {
--- a/src/main/java/org/mitre/fhir/authorization/AuthorizationController.java
+++ b/src/main/java/org/mitre/fhir/authorization/AuthorizationController.java
@@ -84,7 +84,7 @@ public class AuthorizationController {
   @PostMapping(path = "/token", produces = {"application/json"})
   public ResponseEntity<String> getToken(@RequestParam(name = "code", required = false) String code,
       @RequestParam(name = "client_id", required = false) String clientIdRequestParam,
-      @RequestParam(name = "refresh_token", required = false) String refreshToken,
+      @RequestParam(name = "refresh_token", required = false) String refreshTokenValue,
       HttpServletRequest request) throws BearerTokenException {
 
     Log.info("code is " + code);
@@ -99,9 +99,9 @@ public class AuthorizationController {
     if (basicHeader != null) {
       String decodedValue = getDecodedBasicAuthorizationString(basicHeader);
       // client id is user name, and should be before ':'
-      clientId = decodedValue.split(":")[0]; 
+      clientId = decodedValue.split(":")[0];
       // client secret is password, and should be after ':'
-      clientSecret = decodedValue.split(":")[1]; 
+      clientSecret = decodedValue.split(":")[1];
     } else {
       // if no basic auth, client id should be supplied as request param
       clientId = clientIdRequestParam;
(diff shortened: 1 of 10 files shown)


okeefm:
Looks like the Java checkstyle is angry about some stuff in `TokenManager.java`

radamson:
It would be good to include at least a minimal description in the PR

src/main/java/org/mitre/fhir/authorization/AuthorizationController.java
```diff
@@ -116,42 +116,48 @@ public String validateClientId(@PathVariable String clientId, HttpServletRequest
...
       }
-    } catch (TokenNotFoundException tokenNotFoundException) {
-      throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
-          "Refresh Token " + refreshToken + " was not found");
+
+    } else if (refreshTokenValue != null) {
+
+      try {
+        if (refreshTokenValue != null
+            && TokenManager.getInstance().authenticateRefreshToken(refreshTokenValue)) {
```
radamson, line 145:
```suggestion
        if (TokenManager.getInstance().authenticateRefreshToken(refreshTokenValue)) {
```
The condition refreshTokenValue !=null is always true here because of the encapsulating `if`

src/main/java/org/mitre/fhir/authorization/AuthorizationController.java
```diff
@@ -116,42 +116,48 @@ public String validateClientId(@PathVariable String clientId, HttpServletRequest
...
-    }
+      if (fullCode.length >= 3) {
+        String encodedPatientId = fullCode[2];
+        patientId = new String(Base64.getDecoder().decode(encodedPatientId));
+      }
 
-    try {
-      if (refreshToken != null
-          && TokenManager.getInstance().authenticateRefreshToken(actualCodeOrRefreshToken)) {
+      if ((code != null && FhirReferenceServerUtils.SAMPLE_CODE.equals(actualCodeOrRefreshToken))) {
```
radamson, line 137:
```suggestion
      if (FhirReferenceServerUtils.SAMPLE_CODE.equals(actualCodeOrRefreshToken)) {
```

`code != null` is always true here.

src/main/java/org/mitre/fhir/authorization/AuthorizationController.java
```diff
@@ -116,42 +116,48 @@ public String validateClientId(@PathVariable String clientId, HttpServletRequest
...
+          String refreshTokenScopes = refreshToken.getScopesString();
+          return generateBearerTokenResponse(request, clientId, refreshTokenScopes, patientId);
+        }
+      } catch (TokenNotFoundException tokenNotFoundException) {
+        throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
+            "Refresh Token " + refreshTokenValue + " was not found");
+      }
+    } else {
+      throw new ResponseStatusException(HttpStatus.BAD_REQUEST, "Invalid code");
     }
```
radamson, line 157:
```suggestion
    }
```

this else looks unecessary because the line below will throw the exception anyway.

src/main/java/org/mitre/fhir/authorization/FakeOauth2AuthorizationInterceptorAdaptor.java
```diff
@@ -37,23 +34,18 @@ public boolean incomingRequestPostProcessed(RequestDetails requestDetails,
...
+    } catch (TokenNotFoundException tokenNotFoundException) {
       throw new InvalidBearerTokenException(bearerToken);
     }
 
-    String encodedScopes = splitBearerTokenParts[1];
-
-    String scopes = new String(Base64.decode(encodedScopes));
-
-    List<String> scopesArray = Arrays.asList(scopes.split(" "));
+    // List<String> scopesArray = Arrays.asList(scopes.split(" "));
```
radamson, line 48:
```suggestion
```

src/main/java/org/mitre/fhir/authorization/token/Token.java
```diff
@@ -1,16 +1,24 @@
 package org.mitre.fhir.authorization.token;
 
+import java.util.List;
 import java.util.UUID;
 
 public class Token {
 
   private boolean active = true;
+  private List<String> scopes;
```
radamson, line 9:
```suggestion
  private final List<String> scopes;
```

src/main/java/org/mitre/fhir/authorization/token/Token.java
```diff
@@ -1,16 +1,24 @@
...
+  private String patientId;
 
   private String tokenValue;
```
radamson, line 12:
```suggestion
  private final String tokenValue;
```

src/main/java/org/mitre/fhir/authorization/token/Token.java
```diff
@@ -25,4 +33,32 @@ public String getTokenValue() {
...
+    String scopesString = "";
+    for (String scope : scopes) {
+      scopesString += scope + " ";
+    }
+
+    scopesString = scopesString.strip();
+
+    return scopesString;
+
+  }
```
radamson, line 54: I think we might want to use `StringBuilder` here.  The compiler might optimize this anyways, but it seems more appropriate regardless.

```suggestion
    StringBuilder scopesString = new StringBuilder();
    for (String scope : scopes) {
      scopesString.append(scope).append(" ");
    }

    return scopesString.toString().strip();
```

src/main/java/org/mitre/fhir/authorization/AuthorizationController.java
```diff
@@ -215,7 +215,6 @@ private String generateBearerToken(HttpServletRequest request, String clientId,
...
 
     // get their id
```
okeefm, line 217: Would it be worth removing this comment, since the commented-out code below it was removed?

src/main/java/org/mitre/fhir/authorization/AuthorizationController.java
```diff
@@ -116,42 +116,45 @@ public String validateClientId(@PathVariable String clientId, HttpServletRequest
 
     if (code != null) {
       fullCodeString = code;
-    } else if (refreshToken != null) {
-      fullCodeString = refreshToken;
-    } else {
-      throw new ResponseStatusException(HttpStatus.BAD_REQUEST, "Invalid code");
-    }
```
okeefm, line 119: For formatting's sake, should probably get rid of this extra space here
```suggestion
```

src/main/java/org/mitre/fhir/authorization/AuthorizationController.java
okeefm, line 217: This comment can probably get removed, seeing as the code to get the patient ID is removed below as well.
```suggestion
```

src/main/java/org/mitre/fhir/authorization/AuthorizationController.java
```diff
@@ -116,42 +116,45 @@ public String validateClientId(@PathVariable String clientId, HttpServletRequest
...
 
-    // the provided code is actualcode.scopes
-    String[] fullCode = fullCodeString.split("\\.");
 
-    String actualCodeOrRefreshToken = fullCode[0];
+      // the provided code is actualcode.scopes
```
okeefm, line 121: What is `actualcode.scopes` here? I can't see that variable.

src/main/java/org/mitre/fhir/authorization/AuthorizationController.java
```diff
@@ -116,42 +116,45 @@ public String validateClientId(@PathVariable String clientId, HttpServletRequest
...
+          return generateBearerTokenResponse(request, clientId, refreshTokenScopes, patientId);
+        }
+      } catch (TokenNotFoundException tokenNotFoundException) {
+        throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
+            "Refresh Token " + refreshTokenValue + " was not found");
+      }
     }
 
+
+
```
okeefm, line 157: Just another suggestion to remove some extraneous newlines.

```suggestion
```

src/main/java/org/mitre/fhir/authorization/AuthorizationController.java
```diff
@@ -214,8 +210,6 @@ private String generateBearerToken(HttpServletRequest request, String clientId,
       throw new ResponseStatusException(HttpStatus.UNAUTHORIZED, "No patients found");
     }
 
-    // get their id
-    // String patientId = patient.getIdElement().getIdPart();
     List<String> scopesList = Arrays.asList(scopes.split(" "));
```
radamson, line 213: This should be split on something like:

```suggestion
    List<String> scopesList = Arrays.asList(scopes.split(" +"));
```

So that it works when multiple spaces are between scopes.

e.g. doing using this scope:

`launch/patient openid fhirUser offline_access patient/Medication.read patient/Condition.read  patient/Observation.read patient/Patient.read` which has 2 spaces between the Condition and Observation scopes

gives:

Screen Shot 2020-09-08 at 1 46 46 PM

radamson: Also may need to be updated here too as @okeefm pointed out:

https://github.com/inferno-community/inferno-reference-server/pull/17/files#diff-ad5676917be3354a5199f68027f4e7b0R43

radamson (approved):
LGTM.  Tested and everything passed as expected.

src/main/java/org/mitre/fhir/authorization/AuthorizationController.java
radamson: Still seeing the issue even with the updated commits.

I ran `docker-compose up --build` and made sure to pull the changes.  Should I be doing something else?

hershilpatel: oh I realize that maybe the regex is still not extracting each of the spaces , like it it being timid. Do you mind if I move this to another small ticket, where I'll handle splitting the scope to and from an array in a separate method?

radamson: That works.  Can you add a reference to the ticket number here once its created?

hershilpatel:
The scopes splitting over space issue will be handled in https://oncprojectracking.healthit.gov/support/browse/FI-1000

okeefm (approved):

PR #19, opened 2020-07-29
updated readme to specify java version

hershilpatel:
Updated the README to specify the requirement to run the reference server with java 11 if running it manually

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,14 +24,14 @@ The database will be initially populated by the default initdb.sql script. To up
 
 ## Running without Docker
 
+**System Requirements:**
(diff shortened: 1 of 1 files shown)


radamson:
Instead of a `Note` I would make it bold or as a subsection titled `System Requirements`.  Also, I would clarify that it requires Java 1.8 or above.

See this [example from Synthea](https://github.com/synthetichealth/synthea#installation) and [the `validator-wrapper`](https://github.com/inferno-community/fhir-validator-wrapper#installation).  I think it just looks a little cleaner and consistent.

Also please add a minimal description to this PR for posterity.

PR #21, opened 2020-09-22
updated ui to look nicer and have a banner at the top to distinguish reference server from inferno

hershilpatel:
addessing issues

https://oncprojectracking.healthit.gov/support/browse/FI-786

and 

https://oncprojectracking.healthit.gov/support/browse/FI-879

applied some basic twitter boostrap css


okeefm (approved):
Changes look good to me.

src/main/webapp/WEB-INF/templates/patient-picker.html
```diff
@@ -6,22 +6,39 @@
...
+	    <th align="left">ID</th>
+	    <th align="left">Name</th>
+	    <th align="left">Select</th>
+	    </thead>
+	</table>
+</div>
 
 <script src="js/lib/jquery-3.4.1/jquery-3.4.1.min.js"></script>
 <script src="js/lib/DataTables-1.10.20/js/jquery.dataTables.min.js"></script>
+<script src="https://cdn.datatables.net/1.10.22/js/dataTables.bootstrap4.min.js"></script>
```
arscan, line 35: It is weird to me that we are serving some libraries from a cdn and some from our webserver.  Is there a reason why are are not being consistent?

arscan:
I see you updated one of the references to be local, but there are a ton more, so we are still being inconsistent.  Search for `https` on that page.  Also, remove the commented out lines (if you somehow want them back, you could always go into the git log).

arscan (approved):
Looks good!

PR #22, opened 2020-10-01
FI 895 uscore reference must support patients

okeefm:
This Pull Request replaces the Reference Server patients with ones that satisfy FI-895, FI-979, and FI-1003.


hershilpatel:
Ran it and the test still pass

hershilpatel (approved):

arscan:
Do you need to wipe the DB or something first?

arscan (approved):
Wiped it and reran it and everything passed with the two new patients on development (though if you now go to HEAD of development it will skip again because of the requirement for choice types).

We need to remember to update all presets to use the new patientids, everywhere!  Please help make sure I do not forget about this @okeefm.

This will have the unfortunate side-effect of people using an old version of inferno locally that points to the inferno.healthit.gov reference server (which is what the localhost preset does) have the wrong patientids listed, but there isn't much we can do about that.

PR #24, opened 2020-10-06
Fi 1036 rebase

hershilpatel:
Redoing `added checking for skip token auth environment var` pr https://github.com/inferno-community/inferno-reference-server/pull/23 to handle rebase problems, should be working now


okeefm (approved):
Tested this branch; still looks good.

arscan:
So "out of the box" this should fail Inferno Program Edition tests, which is fine because it'll fail SSL as well.  But just something we need to be aware of.

arscan (approved):

PR #25, opened 2020-10-12
added support for specifying a custom port in an env variable

hershilpatel:
This is due to the nginx issue where for some reason the port is on 8443 instead of 443.  

To test this, build this version of the reference server, and run it in site overlay with branch https://github.com/onc-healthit/inferno-site-overlay/tree/custom-port-quick-fix

Go to 
https://localhost/reference-server/r4/.well-known/smart-configuration

and confirm that token_endpoint does not have the port in it

### DIFF
--- a/src/main/java/org/mitre/fhir/utils/FhirReferenceServerUtils.java
+++ b/src/main/java/org/mitre/fhir/utils/FhirReferenceServerUtils.java
@@ -20,6 +20,8 @@ public class FhirReferenceServerUtils {
 
   public static final String DEFAULT_SCOPE = "launch/patient patient/*";
 
+  public static final String CUSTOM_PORT_KEY = "CUSTOM_PORT";
+  
   private static final String HTTP = "http";
   private static final String HTTPS = "https";
   private static final int HTTP_DEFAULT_PORT = 80;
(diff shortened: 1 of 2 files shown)


src/main/java/org/mitre/fhir/utils/FhirReferenceServerUtils.java
```diff
@@ -36,9 +39,19 @@ public static String getServerBaseUrl(HttpServletRequest request) {
     int portNumber = request.getServerPort();
     String port = ":" + portNumber;
 
+    String customPortString = System.getenv(CUSTOM_PORT_KEY);
+
+    int customPortNumber = -1;
+    try {
+      customPortNumber = Integer.parseInt(customPortString);
+    } catch (NumberFormatException numberFormatException) {
+      Log.info("No curstom port found");
```
Jammjammjamm, line 48:
```suggestion
      Log.info("No custom port found");
```

src/main/java/org/mitre/fhir/utils/FhirReferenceServerUtils.java
```diff
@@ -20,6 +21,8 @@
 
   public static final String DEFAULT_SCOPE = "launch/patient patient/*";
 
+  public static final String CUSTOM_PORT_KEY = "CUSTOM_PORT";
```
Jammjammjamm, line 24: I think this needs a clearer name. Or at least an explanatory comment.

hershilpatel: hmm maybe DEPLOYED_PORT? do you have any suggestions

hershilpatel: I think I am going to go with CUSTOM_DEPLOYMENT_PORT, I will also add a comment explaining

Jammjammjamm (approved):

PR #27, opened 2020-10-20
undoing the custom deployment as it is no longer necessary

hershilpatel:
Removes the custom deployment support to avoid unecessary technical debt


arscan (approved):
Checked and it looks like it reverts everything in #25

okeefm (approved):
Looks good.

PR #29, opened 2020-12-01
Landing page

hershilpatel:
Added landing page at localhost/reference-server

Screen Shot 2020-12-02 at 11 06 30 AM


src/main/webapp/WEB-INF/templates/landing.html
```diff
@@ -0,0 +1,56 @@
...
+
+<body>
+	<div th:replace="tmpl-banner :: banner"></div>
+	
+	<br>
+	<br>
+	
+    <div id="pageContent" class="container">
+   		
+   		This server is designed to meet the Inferno Program Certification Tests.
```
Jammjammjamm, line 22: You should use some p tags. HTML doesn't respect whitespace.

src/main/webapp/WEB-INF/templates/landing.html
```diff
@@ -0,0 +1,56 @@
...
+		</ul>
+		
+		Support for US Core v3.1.0, including:
+		<ul>
+			<li>Support for required API interactions</li>
+			<li>Example resources and patients</li>
+		</ul>   
+				
+   		Links:
+   		<div class="card card-body bg-light">
```
Jammjammjamm, line 41: I don't see why this list is styled differently than the other lists.

src/main/webapp/WEB-INF/templates/landing.html
```diff
@@ -0,0 +1,56 @@
...
+   		This server is designed to meet the Inferno Program Certification Tests.
+This includes:
+
+		Support for SMART-on-FHIR, including:
+		<ul>
+			<li>Public and Confidential Clients</li>
+			<li>Standalone and EHR Launch</li>
+			<li>Token Refresh</li>
+			<li>Token Revocation</li>
+			<li>.well-known URL</li>
```
Jammjammjamm, line 31: I don't know that this is worth listing. It's a requirement.

src/main/webapp/WEB-INF/templates/landing.html
hershilpatel: Added p tags

src/main/webapp/WEB-INF/templates/landing.html
hershilpatel: Removed unnecessary list

Jammjammjamm: I just meant the `.well-known URL` entry.

hershilpatel: My mistake, re-adding

src/main/webapp/WEB-INF/templates/landing.html
hershilpatel: Removed different styling

src/main/webapp/WEB-INF/templates/landing.html
```diff
@@ -0,0 +1,52 @@
...
+
+<body>
+	<div th:replace="tmpl-banner :: banner"></div>
+	
+	<br>
+	<br>
+	
+    <div id="pageContent" class="container">
+   		
+   		<p>This server is designed to meet the Inferno Program Certification Tests.  This includes:</p>
```
okeefm, line 22:
```suggestion
   		<p>This server is designed to meet the requirements of the Inferno Program Certification Tests.  This includes:</p>
```

src/main/webapp/WEB-INF/templates/landing.html
```diff
@@ -0,0 +1,52 @@
...
+
+<p>Support for SMART-on-FHIR, including:</p>
+		<ul>
+			<li>Public and Confidential Clients</li>
+			<li>Standalone and EHR Launch</li>
+			<li>Token Refresh</li>
+			<li>Token Revocation</li>
+		</ul>
+
+<p>Support for US Core v3.1.0, including:</p>
```
okeefm, line 32: Should we bump this to `v3.1.1`, since we're about to release 3.1.1 support?
```suggestion
<p>Support for US Core v3.1.1, including:</p>
```

okeefm (commented):
Couple quick suggestions, otherwise, looks good

src/main/webapp/WEB-INF/templates/landing.html
```diff
@@ -0,0 +1,52 @@
+<!doctype html>
+<!-- Normally this will be a site to accept a user name and password and log in to determine which code to send, for now it'll just redirect-->
```
arscan, line 2: ?

src/main/webapp/WEB-INF/templates/landing.html
```diff
@@ -0,0 +1,52 @@
...
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+
+  <title>Inferno Reference Server</title>
+  <meta name="description" content="Authorize Cindehr">
+  
+  <link href="../reference-server/app/js/lib/bootstrap-4.5.3-dist/css/bootstrap.min.css" rel="stylesheet"> 
+  <link href="../reference-server/app/css/site.css" rel="stylesheet">
```
arscan, line 11: why the `../reference-server/`?  Would just `/reference-server/` do?

src/main/webapp/WEB-INF/templates/landing.html
```diff
@@ -0,0 +1,52 @@
+<!doctype html>
+<!-- Normally this will be a site to accept a user name and password and log in to determine which code to send, for now it'll just redirect-->
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+
+  <title>Inferno Reference Server</title>
+  <meta name="description" content="Authorize Cindehr">
```
arscan, line 8: I'm not sure this description makes sense anymore.

src/main/webapp/WEB-INF/templates/landing.html
```diff
@@ -0,0 +1,52 @@
...
+
+<p>Support for US Core v3.1.1, including:</p>
+		<ul>
+			<li>Support for required API interactions</li>
+			<li>Example resources and patients</li>
+		</ul>   
+				
+   		<p>Links:</p>
+   		<ul>
+   			<li><a href="/reference-server/r4/metadata">Conformance</a></li>
```
arscan, line 40: It is now called a 'Capability Statement'.  I think a link to the endpoint itself might be more useful (even though it won't actually do anything in a browser).  I could imagine going to this page and wanting to grab all the config info I would need to run inferno against it. Which would be the fhir endpoint, and the confidential/public client details.

src/main/webapp/WEB-INF/templates/landing.html
hershilpatel: copy and paste error, removing that line

src/main/webapp/WEB-INF/templates/landing.html
hershilpatel: added section  for fhir endpoint, client id, and client secret

arscan: Could you also add the public client id?  So maybe the order is "Public Client ID: XYZ" then "Confidential Client ID: ABC" and "Confidential Client Secret: DEF".

arscan: (ideally these values would be driven from the constant defined in the app also, because it makes me a little uncomfortable to have the same thing repeated in multiple places, but it seems unlikely that we'll mess with these so that isn't a huge deal to me, just mentioning it as more of a best practice)

src/main/java/org/mitre/fhir/utils/FhirReferenceServerUtils.java
```diff
@@ -17,10 +17,10 @@
...
   public static final String REFERENCE_SERVER_PATH = "/app";
 
-  public static final String SAMPLE_PUBLIC_CLIENT_ID = "SAMPLE_PUBLIC_CLIENT_ID";
-  public static final String SAMPLE_CONFIDENTIAL_CLIENT_ID = "SAMPLE_CONFIDENTIAL_CLIENT_ID";
-  public static final String SAMPLE_CONFIDENTIAL_CLIENT_SECRET =
-      "SAMPLE_CONFIDENTIAL_CLIENT_SECRET";
+  //public static final String SAMPLE_PUBLIC_CLIENT_ID = "SAMPLE_PUBLIC_CLIENT_ID";
+  //public static final String SAMPLE_CONFIDENTIAL_CLIENT_ID = "SAMPLE_CONFIDENTIAL_CLIENT_ID";
+  //public static final String SAMPLE_CONFIDENTIAL_CLIENT_SECRET =
+  //    "SAMPLE_CONFIDENTIAL_CLIENT_SECRET";
```
arscan, line 23: Try to get in the habit of reviewing your diffs before committing and deleting commented out code (unless it serves a good purpose, but I'm not seeing what that purpose might be in this case).

arscan (approved):

src/main/java/org/mitre/fhir/HapiReferenceServerProperties.java
```diff
@@ -232,7 +237,38 @@ public String getHibernateSearchLuceneVersion() {
         .getProperty(HIBERNATE_SEARCH_LUCENE_VERSION_KEY);
     return hibernateSearchLuceneVersion;
   }
+  
+  /**
+   * Returns the Hibernate Property.
```
okeefm, line 242: Is this property (and the two below) actually a `Hibernate` property?
```suggestion
   * Returns the public Client ID Property.
```

src/main/java/org/mitre/fhir/HapiReferenceServerProperties.java
```diff
@@ -232,7 +237,38 @@ public String getHibernateSearchLuceneVersion() {
...
+   * @return the property
+   */ 
+  public String getPublicClientId() {
+    String publicClientId = properties
+        .getProperty(PUBLIC_CLIENT_ID_KEY);
+    return publicClientId;
+  }
+
+  /**
+   * Returns the Hibernate Property.
```
okeefm, line 252:
```suggestion
   * Returns the Confidential Client ID Property.
```

src/main/java/org/mitre/fhir/HapiReferenceServerProperties.java
```diff
@@ -232,7 +237,38 @@ public String getHibernateSearchLuceneVersion() {
...
+   * @return the property
+   */ 
+  public String getConfidentialClientId() {
+    String confidentialClientId = properties
+        .getProperty(CONFIDENTIAL_CLIENT_ID_KEY);
+    return confidentialClientId;
+  }
 
+  /**
+   * Returns the Hibernate Property.
```
okeefm, line 262:
```suggestion
   * Returns the Confidential Client Secret Property.
```

okeefm (changes requested):
Just a couple of Javadoc comment changes, otherwise looks good!

okeefm (approved):
Looks good!

PR #32, opened 2020-12-07
quick fix for failing token revocation test on inferno

hershilpatel:
fixes issue where after a token is revoked you can still use it 

fixed issue where submission doesn't display results when you revoke a token


src/main/webapp/WEB-INF/templates/revoke-token.html
```diff
@@ -42,15 +42,36 @@
...
-	        $.post(url, data, function(data, status) {
+	        
+      		/*$.post(url, data, function(data, status) {
 	            
 	            alert("Token successfully revoked.")
 
 	        }).fail(function() {
 	            alert("Error revoking token");
-	        });
+	        });*/
```
arscan, line 56: Do you need this commented out code?

arscan:
Could you confirm that you ran Inferno's revocation test against this successfully?  I am still getting the following, and based on inspection of the html/js I'm pretty sure I am running the most recent version:

Screen Shot 2020-12-11 at 11 20 06 AM

arscan (approved):
Looks great!

PR #33, opened 2020-12-10
made checkbox labels selectable

hershilpatel:
Make the checkboxes easier to select/deselect by allowing the user to click on their lables


arscan:
It works, but you need to update the `?s=v2` string (to `?s=v3`) on authorization.html, otherwise most return visitors won't ever download this code because of our aggressive cache headers.  Make sure you do that any time you update any css or js file.

arscan (approved):

PR #34, opened 2020-12-14
Update error message look

hershilpatel:
Replaces the default html/javascript alert message with error message on the page.  
Screen Shot 2020-12-14 at 8 16 05 AM

Screen Shot 2020-12-14 at 8 17 38 AM


src/main/webapp/WEB-INF/templates/authorization.html
```diff
@@ -16,16 +16,20 @@
...
-
-        <br>
-
-        <div id="scopes" class="well"></div>
-
-        <br>
-
-        <button id="submit" type="button" class="btn btn-info">Authorize</button>
+    <div class="container">
+        <div style="display: none" class="card card-body bg-danger text-white" id="errorMessage"> </div>
```
okeefm, line 20: On the `revoke-token` page this div is inside the `pageContent` div, whereas here it's not. Was there a reason this was placed outside of `pageContent` on this page?

okeefm: Or, conversely, that it was in `pageContent` on the other page

src/main/webapp/WEB-INF/templates/revoke-token.html
```diff
@@ -40,10 +46,11 @@
...
       $(document).ready(function() {
 
-
-      	$('#revokeTokenButton').click(function() {
-      	
-      		console.log("Revoke Button Pressed");
+      	$('#revokeTokenButton').click(function() {      	
+			$('#errorMessage').hide();
+			$('#errorMessage').removeClass('bg-danger');
+			$('#errorMessage').removeClass('bg-info');
```
okeefm, line 52: I think you could, in theory, chain these methods (and other places that these get called in series), so jQuery doesn't have to do the selector 3 times in a row.
```suggestion
			$('#errorMessage').hide();
			                               .removeClass('bg-danger');
			                               .removeClass('bg-info');
```

src/main/webapp/WEB-INF/templates/authorization.html
hershilpatel: updated to be consistent (error message outside of pageContent)

src/main/webapp/WEB-INF/templates/revoke-token.html
hershilpatel: Chained it in multiple places in new commit!

okeefm (approved):
Looks good!

arscan (approved):
Looks good.  I'm having an issue revoking a token, which we can address separately.  It could be because this is on an older code branch and hasn't been kept up to date with master?

PR #35, opened 2021-01-06
Fi 1093 change skip auth to specifiy token value

hershilpatel:
Removing the option to Skip Authentication and instead adding a env var to contain a custom bearer token.

To Test, set an env var CUSTOM_BEARER_TOKEN to value and try to enact a request with that bearer token.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,9 +45,9 @@ To use as a confidential client, use `SAMPLE_CONFIDENTIAL_CLIENT_ID` as the clie
 
 To launch an app from the EHR go to `reference-server/app/app-launch`
 
-## Run without an Authentication Token
+## Custom Authentication Token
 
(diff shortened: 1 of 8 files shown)


okeefm:
I think if this PR is rebased on `master`, it should rerun with Github Actions rather than Travis (which should pass)

okeefm (commented):
Does what it should! Just two quick notes:

* It looks like some files that are already on `master` are getting re-added in this PR (specifically the GitHub Actions workflow files). It may be worth backing out the last merge commit and rebasing the entire branch on `master`?
* We should probably remove the old `SKIP_TOKEN_AUTHENTICATION` variable in `docker-compose.yml`, and possibly swap it out with a `CUSTOM_BEARER_TOKEN` (maybe something like `SAMPLE_TOKEN`)?

okeefm (approved):
Tested, looks good!

arscan (approved):
Works great!

PR #37, opened 2021-01-28
Fi 1116 update version of hapi to 5.3

hershilpatel:
Updating the version of hapi to version 5.0 (and updating the code to still work).

To test, confirm that the unit tests pass, and confirm that the reference server still passes inferno tests


arscan:
Should our POM have something other than 1.0.0-SNAPSHOT as our version? I would assume this would accompany a new version but I don't really know when that change would get introduced into that file.

arscan:
I'm getting a failure on our patient name + birthdate search (USCP-04).  I think this is fixed in v5.3.0, because their birthdate search works on their deployed version (hapi.fhir.org/baseR4)... perhaps in [this PR](https://github.com/hapifhir/hapi-fhir/pull/2302).  I recommend *not* merging into master as an in-between to the latest version of HAPI, because it fails our tests.  We should just go up to v5.3.

arscan (approved):
Looks good, and tests all still pass.  Also, this speeds up the time it takes to run all our tests by 40%!

arscan:
@hershilpatel -- could you open a Draft PR to put the updated `initdb.sql` in https://github.com/onc-healthit/inferno.healthit.gov?

hershilpatel:
> @hershilpatel -- could you open a Draft PR to put the updated `initdb.sql` in https://github.com/onc-healthit/inferno.healthit.gov?

Added https://github.com/onc-healthit/inferno.healthit.gov/pull/77

okeefm (approved):
Tested, looks good

PR #38, opened 2021-03-04
FI-1112 Fix CapabilityStatement validation errors

hershilpatel:
Value was already hard coded into reference server, just needed to correct it to what seems like the correct value

https://www.hl7.org/fhir/capabilitystatement-definitions.html#CapabilityStatement.rest.security

https://www.hl7.org/fhir/valueset-restful-security-service.html


arscan:
I'm still getting this error, which is what I thought you were addressing:

Screen Shot 2021-03-09 at 12 27 00 PM

arscan (approved):
Removing the operations fixes the issues.  Tests now pass:

Screen Shot 2021-03-09 at 2 45 45 PM

arscan:
@hershilpatel please make a new ticket for removing all non-uscdi resources to trim this down a bit further, as well as address the warnings in USCCAP-05.

PR #39, opened 2021-03-10
updated version to 1.0.1 for release

hershilpatel:


pom.xml
```diff
@@ -4,7 +4,7 @@
 
     <groupId>org.mitre.fhir</groupId>
     <artifactId>inferno-reference-server</artifactId>
-    <version>1.0.0-SNAPSHOT</version>
+    <version>1.0.1</version>
```
arscan, line 7: Should we be v1.0.2 now?  We already have a 1.0.1 release in github, so things got a bit out of synch it looks.

hershilpatel: Good catch, to avoid confusion in branch names, I am going to close this one, update https://github.com/inferno-community/inferno-reference-server/pull/40 to not be snapshot, and add a pr for 1.0.3-SNAPSHOT

PR #43, opened 2021-03-26
Fi 1178 capability statement from scratch

hershilpatel:
Added code to generate a simple capability statement that only has elements relevant to passing inferno


src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
```diff
@@ -90,32 +109,324 @@ public CapabilityStatement getServerConformance(HttpServletRequest theRequest,
...
+   * 
+   * @return List of Capability Statement resources with expected attributes for Inferno Testing
+   */
+  private List<CapabilityStatementRestResourceComponent> getResources() {
+
+    CapabilityStatementRestResourceComponent allergyIntolerance =
+        new CapabilityStatementRestResourceComponent();
+    allergyIntolerance.setType(ALLERGY_INTOLERANCE);
+    allergyIntolerance.setProfile(getProfile(ALLERGY_INTOLERANCE));
+    allergyIntolerance.addSupportedProfile(getProfile("us-core-allergyintolerance"));
```
yunwwang, line 126: From reading the code, I don't think the url for supportedProfile is correct. The url for us-core-allergyintolerance is: http://hl7.org/fhir/us/core/StructureDefinition/us-core-allergyintolerance. The result from getProfile looks like to be: "http://hl7.org/fhir/StructureDefinition/us-core-allergyintolerance"

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
```diff
@@ -90,32 +109,324 @@ public CapabilityStatement getServerConformance(HttpServletRequest theRequest,
...
+
+  private static String getProfile(String profileName) {
+    return PROFILE_PREFIX + "/" + profileName + "/";
+  }
+
+  @Override
+  public CapabilityStatement getServerConformance(HttpServletRequest theRequest,
+      RequestDetails theRequestDetails) {
+
+    CapabilityStatement capabilityStatement = new CapabilityStatement();
```
yunwwang, line 406: An overall question. Why not creating CapabilityStatement as a json/xml file then parsing the file here. Like:
```
CapabilityStatement capabilityStatement = parser.parserResource(CapabilityStatement.class, input)
```

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
```diff
@@ -90,32 +109,324 @@ public CapabilityStatement getServerConformance(HttpServletRequest theRequest,
...
+    condition.addInteraction().setCode(TypeRestfulInteraction.SEARCHTYPE);
+    resources.add(condition);
+
+    CapabilityStatementRestResourceComponent device =
+        new CapabilityStatementRestResourceComponent();
+    device.setType(DEVICE);
+    device.setProfile(getProfile(DEVICE));
+    device.addSupportedProfile(getProfile("us-core-implantable-device"));
+    device.addSearchParam().setName("patient").setType(SearchParamType.REFERENCE);
+    device.addSearchParam().setName("category").setType(SearchParamType.TOKEN);
```
yunwwang, line 180: Where are these three search parameters defined? "category", "date", "code"

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
```diff
@@ -90,32 +109,324 @@ public CapabilityStatement getServerConformance(HttpServletRequest theRequest,
...
+    diagnosticReport.addInteraction().setCode(TypeRestfulInteraction.SEARCHTYPE);
+    resources.add(diagnosticReport);
+
+    CapabilityStatementRestResourceComponent documentReference =
+        new CapabilityStatementRestResourceComponent();
+    documentReference.setType(DOCUMENT_REFERENCE);
+    documentReference.setProfile(getProfile(DOCUMENT_REFERENCE));
+    documentReference.addSupportedProfile(getProfile("us-core-documentreference"));
+    documentReference.addSearchParam().setName("patient").setType(SearchParamType.REFERENCE);
+    documentReference.addSearchParam().setName("_id").setType(SearchParamType.TOKEN);
```
yunwwang, line 211: The "_id" search parameter is not consistently presented cross resources. Should we add "_id" search param to all resources

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
```diff
@@ -90,32 +109,324 @@ public CapabilityStatement getServerConformance(HttpServletRequest theRequest,
...
+    documentReference.addSearchParam().setName("type").setType(SearchParamType.TOKEN);
+    documentReference.addSearchParam().setName("category").setType(SearchParamType.TOKEN);
+    documentReference.addSearchParam().setName("date").setType(SearchParamType.DATE);
+    documentReference.addSearchParam().setName("code").setType(SearchParamType.TOKEN);
```
yunwwang, line 215: DiagnosticReport does not have "code" search parameter

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
```diff
@@ -90,32 +109,324 @@ public CapabilityStatement getServerConformance(HttpServletRequest theRequest,
...
+    documentReference.addInteraction().setCode(TypeRestfulInteraction.SEARCHTYPE);
+    resources.add(documentReference);
+
+    CapabilityStatementRestResourceComponent encounter =
+        new CapabilityStatementRestResourceComponent();
+    encounter.setType(ENCOUNTER);
+    encounter.setProfile(getProfile(ENCOUNTER));
+    encounter.addSupportedProfile(getProfile("us-core-encounter"));
+    encounter.addSearchRevInclude(SEARCH_REV_INCLUDE);
+    encounter.addInteraction().setCode(TypeRestfulInteraction.READ);
```
yunwwang, line 228: Need search parameters for Encounter

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
```diff
@@ -90,32 +109,324 @@ public CapabilityStatement getServerConformance(HttpServletRequest theRequest,
...
+    immunization.addInteraction().setCode(TypeRestfulInteraction.VREAD);
+    immunization.addInteraction().setCode(TypeRestfulInteraction.SEARCHTYPE);
+    resources.add(immunization);
+
+    CapabilityStatementRestResourceComponent location =
+        new CapabilityStatementRestResourceComponent();
+    location.setType(LOCATION);
+    location.setProfile(getProfile(LOCATION));
+    location.addSupportedProfile(getProfile("us-core-location"));
+    location.addSearchParam().setName("near").setType(SearchParamType.SPECIAL);
```
yunwwang, line 261: should this be "address" with type "string"?

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
```diff
@@ -90,32 +109,324 @@ public CapabilityStatement getServerConformance(HttpServletRequest theRequest,
...
+
+    CapabilityStatementRestResourceComponent medication =
+        new CapabilityStatementRestResourceComponent();
+    medication.setType(MEDICATION);
+    medication.setProfile(getProfile(MEDICATION));
+    medication.addSupportedProfile(getProfile("us-core-medication"));
+    medication.addSearchRevInclude(SEARCH_REV_INCLUDE);
+    medication.addInteraction().setCode(TypeRestfulInteraction.READ);
+    medication.addInteraction().setCode(TypeRestfulInteraction.VREAD);
+    medication.addInteraction().setCode(TypeRestfulInteraction.SEARCHTYPE);
```
yunwwang, line 276: SearchType is not "required" for Medicaiton

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
```diff
@@ -90,32 +109,324 @@ public CapabilityStatement getServerConformance(HttpServletRequest theRequest,
...
+    location.addInteraction().setCode(TypeRestfulInteraction.VREAD);
+    location.addInteraction().setCode(TypeRestfulInteraction.SEARCHTYPE);
+    resources.add(location);
+
+    CapabilityStatementRestResourceComponent medication =
+        new CapabilityStatementRestResourceComponent();
+    medication.setType(MEDICATION);
+    medication.setProfile(getProfile(MEDICATION));
+    medication.addSupportedProfile(getProfile("us-core-medication"));
+    medication.addSearchRevInclude(SEARCH_REV_INCLUDE);
```
yunwwang, line 273: Medication does not support _revInclude

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
```diff
@@ -90,32 +109,324 @@ public CapabilityStatement getServerConformance(HttpServletRequest theRequest,
...
+    immunization.addInteraction().setCode(TypeRestfulInteraction.SEARCHTYPE);
+    resources.add(immunization);
+
+    CapabilityStatementRestResourceComponent location =
+        new CapabilityStatementRestResourceComponent();
+    location.setType(LOCATION);
+    location.setProfile(getProfile(LOCATION));
+    location.addSupportedProfile(getProfile("us-core-location"));
+    location.addSearchParam().setName("near").setType(SearchParamType.SPECIAL);
+    location.addSearchRevInclude(SEARCH_REV_INCLUDE);
```
yunwwang, line 262: Location does not support _revInclude

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
```diff
@@ -90,32 +109,324 @@ public CapabilityStatement getServerConformance(HttpServletRequest theRequest,
...
+    observation.addInteraction().setCode(TypeRestfulInteraction.VREAD);
+    observation.addInteraction().setCode(TypeRestfulInteraction.SEARCHTYPE);
+    resources.add(observation);
+
+    CapabilityStatementRestResourceComponent organization =
+        new CapabilityStatementRestResourceComponent();
+    organization.setType(ORGANIZATION);
+    organization.setProfile(getProfile(ORGANIZATION));
+    organization.addSupportedProfile(getProfile("us-core-organization"));
+    organization.addSearchRevInclude(SEARCH_REV_INCLUDE);
```
yunwwang, line 325: Organization does not support _revInclude

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
```diff
@@ -90,32 +109,324 @@ public CapabilityStatement getServerConformance(HttpServletRequest theRequest,
...
+    organization.addInteraction().setCode(TypeRestfulInteraction.READ);
+    organization.addInteraction().setCode(TypeRestfulInteraction.VREAD);
+    organization.addInteraction().setCode(TypeRestfulInteraction.SEARCHTYPE);
```
yunwwang, line 328: Organization shall support search by "name" and "address"

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
```diff
@@ -90,32 +109,324 @@ public CapabilityStatement getServerConformance(HttpServletRequest theRequest,
...
+    practitioner.addInteraction().setCode(TypeRestfulInteraction.VREAD);
+    practitioner.addInteraction().setCode(TypeRestfulInteraction.SEARCHTYPE);
+    resources.add(practitioner);
+
+    CapabilityStatementRestResourceComponent practitionerRole =
+        new CapabilityStatementRestResourceComponent();
+    practitionerRole.setType(PRACTITIONER_ROLE);
+    practitionerRole.setProfile(getProfile(PRACTITIONER_ROLE));
+    practitionerRole.addSupportedProfile(getProfile("us-core-practitionerrole"));
+    practitionerRole.addSearchRevInclude(SEARCH_REV_INCLUDE);
```
yunwwang, line 365: PractitionerRole does not support _revInclude

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
```diff
@@ -90,32 +109,324 @@ public CapabilityStatement getServerConformance(HttpServletRequest theRequest,
...
+    practitionerRole.addInteraction().setCode(TypeRestfulInteraction.READ);
+    practitionerRole.addInteraction().setCode(TypeRestfulInteraction.VREAD);
+    practitionerRole.addInteraction().setCode(TypeRestfulInteraction.SEARCHTYPE);
```
yunwwang, line 368: PractitionerRole shall support search by "specialty" and "parctitioner"

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
```diff
@@ -90,32 +109,324 @@ public CapabilityStatement getServerConformance(HttpServletRequest theRequest,
...
+    procedure.addInteraction().setCode(TypeRestfulInteraction.VREAD);
+    procedure.addInteraction().setCode(TypeRestfulInteraction.SEARCHTYPE);
+    resources.add(procedure);
+
+    CapabilityStatementRestResourceComponent provenance =
+        new CapabilityStatementRestResourceComponent();
+    provenance.setType(PROVENANCE);
+    provenance.setProfile(getProfile(PROVENANCE));
+    provenance.addSupportedProfile(getProfile("us-core-provenance"));
+    provenance.addSearchRevInclude(SEARCH_REV_INCLUDE);
```
yunwwang, line 389: Provenance does not support _revInclude

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
```diff
@@ -90,32 +109,324 @@ public CapabilityStatement getServerConformance(HttpServletRequest theRequest,
...
+    provenance.addInteraction().setCode(TypeRestfulInteraction.READ);
+    provenance.addInteraction().setCode(TypeRestfulInteraction.VREAD);
+    provenance.addInteraction().setCode(TypeRestfulInteraction.SEARCHTYPE);
```
yunwwang, line 392: Provenance does not support search

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
```diff
@@ -90,32 +109,324 @@ public CapabilityStatement getServerConformance(HttpServletRequest theRequest,
...
+    CapabilityStatementImplementationComponent implementation =
+        new CapabilityStatementImplementationComponent();
+    implementation.setDescription("HAPI FHIR R4 Server");
+    implementation.setUrl(FhirReferenceServerUtils.getServerBaseUrl(theRequest));
+    capabilityStatement.setImplementation(implementation);
+
+    capabilityStatement.setFhirVersion(FHIRVersion._4_0_1);
 
+    capabilityStatement.addFormat("application/fhir+xml");
+    capabilityStatement.addFormat("application/fhir+json");
```
yunwwang, line 421: It would be better to add these elements:
publisher: MITRE
implmentationGuide: http://hl7.org/fhir/us/core/ImplementationGuide/hl7.fhir.us.core|3.1.1

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
hershilpatel: That's a good point, I'll look into it

hershilpatel: So this is possible to do, but the urls need to be adapted to the server's base url anyway, so I think it is easier to do it all in Java rather than read JSON and edit it anyway...

yunwwang: If it is only the server's base url, you can update the attribute after parsing xml/json to CapabilityStatement instance

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
hershilpatel: Fixed https://github.com/inferno-community/inferno-reference-server/pull/43/files#diff-d6ca14b06f93c2efe4d864af27c3166538f5a3d6c79980cd17c3801ab3086b1bR44

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
hershilpatel: Updated to correct search params

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
hershilpatel: Updated to correct search params

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
hershilpatel: Updated to correct search params

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
hershilpatel: Updated to correct search params

hershilpatel:
@okeefm I am sure there I can do something like that, I'll look into making it more concise.

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
hershilpatel: Removed _revInclude

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
hershilpatel: Removed _revInclude

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
```diff
@@ -90,32 +109,324 @@ public CapabilityStatement getServerConformance(HttpServletRequest theRequest,
...
+    resources.add(patient);
+
+    CapabilityStatementRestResourceComponent practitioner =
+        new CapabilityStatementRestResourceComponent();
+    practitioner.setType(PRACTITIONER);
+    practitioner.setProfile(getProfile(PRACTITIONER));
+    practitioner.addSupportedProfile(getProfile("us-core-practitioner"));
+    practitioner.addSearchParam().setName("patient").setType(SearchParamType.REFERENCE);
+    practitioner.addSearchParam().setName("date").setType(SearchParamType.DATE);
+    practitioner.addSearchRevInclude(SEARCH_REV_INCLUDE);
```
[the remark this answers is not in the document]
hershilpatel: removed _revInclude

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
hershilpatel: removed _revInclude

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
hershilpatel: Added searchParams

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
hershilpatel: Removed

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
hershilpatel: Removed Search

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
hershilpatel: Added

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
```diff
@@ -90,32 +109,324 @@ public CapabilityStatement getServerConformance(HttpServletRequest theRequest,
...
+      RequestDetails theRequestDetails) {
+
+    CapabilityStatement capabilityStatement = new CapabilityStatement();
+    capabilityStatement.setStatus(PublicationStatus.ACTIVE);
+    capabilityStatement.setDate(new Date());
+    capabilityStatement.setKind(CapabilityStatementKind.INSTANCE);
+
+    CapabilityStatementImplementationComponent implementation =
+        new CapabilityStatementImplementationComponent();
+    implementation.setDescription("HAPI FHIR R4 Server");
```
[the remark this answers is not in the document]
hershilpatel: changed text

hershilpatel:
@okeefm I looked at the search param further, and there isn't necessarily exact overlap between the different search params resources use. For example, although a lot of them use name, not all of them do.  Although the current code may seem a bit verbose, I think it works at keeping all required info in one place.

yunwwang (approved):

src/main/java/org/mitre/fhir/authorization/ServerConformanceWithAuthorizationProvider.java
arscan: I'd generally agree with @yunwwang on this stylistically, that is a ton of repetitive code modifying the state of an object when it could just be a file using a standard format read from disk (you'd just want to be a little careful that it was read once startup, not on every request) ...and then update just the parts that are dynamic (like the base url) as needed.

arscan:
There's a chance that I misconfigured my build, but it still looks like we are getting lots of warnings on profiles not being supported, and in looking at what is in the generated capability statement, this doesn't look right:

Screen Shot 2021-06-14 at 4 45 58 PM


(at the very least, i think the double "/" in the URLs will mess up our checking... and why the different URLs for what look to be the same profile?)


Let me know if I'm running an old version though.  @hershilpatel

arscan:
You can remove "resource.profile" too (or change it to be consistent with "resource.supportedProfile").  We don't check for that... (I think we may have decided that was too much of a stretch given the definition of that element in our situation).

hershilpatel:
I adding a lot of the missing supported profiles but USCCAP-05 still has warnings because US Core Observation doesn't support bp, bodyheight, bodyweight, heartrate, resprate, bodytemp...I can add these, but I don't think they are required according to https://www.hl7.org/fhir/us/core/CapabilityStatement-us-core-server.html#behavior.  Am Iooking at the right place?

arscan:
Thanks for double checking this.  These are special cases because they are profiles that come from the base fhir specification, and not us core (note the URL for teh StructureDefinitions do not have "/us/core/" in the middle).  However, US core points to them as requirements when in the context of USCDI, which we are in.  See: https://www.hl7.org/fhir/us/core/general-guidance.html#us-core-data-for-interoperability-and-2015-edition-common-clinical-data-set

I do wonder if the US Core Capability Statement should list these as supportedProfiles.  They aren't defined by US Core, but I think they are required to be supported (as much as the other profiles).

Please cross check our list with the Observation profiles from the base fhir specification as another check that we got all this right.

FYI @Jammjammjamm

arscan (approved):
Looks good, no more warnings (the TLS error is expected).

I think using the file-based approach was a good idea long-term. Thanks!

Screen Shot 2021-06-21 at 2 56 41 PM

PR #44, opened 2021-05-11
Add GitHub ci to run inferno tests

hershilpatel:


.github/workflows/test-with-inferno.yml
```diff
@@ -0,0 +1,30 @@
...
+      uses: jakejarvis/wait-action@master
+      with:
+        time: '300s'
+
+    - name: Run Inferno Program tests on Reference Server
+      run:  docker-compose -f ./.github/inferno/docker-compose.yml exec -T inferno_program bundle exec rake inferno:execute[http://referenceserver:8080/reference-server/r4,onc_program,UsCoreR4CapabilityStatement,USCore311Patient,USCore311Allergyintolerance,USCore311Careplan,USCore311Careteam,USCore311Condition,USCore311ImplantableDevice,USCore311DiagnosticreportNote,USCore311DiagnosticreportLab,USCore311Documentreference,USCore311Goal,USCore311Immunization,USCore311Medicationrequest,USCore311Smokingstatus,USCore311PediatricWeightForHeight,USCore311ObservationLab,USCore311PediatricBmiForAge,USCore311PulseOximetry,USCore311HeadOccipitalFrontalCircumferencePercentile,USCore311Bodyheight,USCore311Bodytemp,USCore311Bp,USCore311Bodyweight,USCore311Heartrate,USCore311Resprate,USCore311Procedure,USCoreR4ClinicalNotes,USCore311Encounter,USCore311Location,USCore311Organization,USCore311Practitioner,USCore311Practitionerrole,USCore311Provenance,USCoreR4DataAbsentReason] --  --TOKEN SAMPLE_TOKEN --PATIENT_IDS 85,355 --DEVICE_CODES ""
```
arscan, line 30: Instead of doing three steps `docker-compose up; wait 300; docker-compose exec`, have you considered doing `docker-compose run` instead?  That's how I did it in the proof of concept I believe I pointed you to which seems a lot simpler and more stable (what if it takes over 300 seconds to build?): https://github.com/arscan/smart-scheduling-links/blob/ef7c291f565fb72c6c426a217fcc019e4a3ea6d0/.github/workflows/validate.yml

Or is there something different here preventing us from doing that?

arscan: If you do this because you need to wait for the reference server to start, you could probably add a `depends_on` to the reference server in the docker-compose file, and then wrap the inferno call in some kind of "wait" wrapper script as described here: https://docs.docker.com/compose/startup-order/

.github/inferno/docker-compose.yml
```diff
@@ -0,0 +1,36 @@
...
+    build:
+      context: ../../
+      dockerfile: Dockerfile.database
+    ports:
+      - '5432:5432'
+    volumes:
+      - fhir-pgdata:/var/lib/postgresql/data
+    environment:
+      - POSTGRES_HOST_AUTH_METHOD=trust
+  referenceserver:
```
arscan, line 23: please try to use consistent naming with what is already in here (change to `reference_server`).

hershilpatel:
@arscan I have updated this to use wait-for-it and docker-compose run

okeefm:
This is pretty cool! I wonder if there's a way to separate out the "build" log data, that I believe is coming from Gradle, from the output of the test runs. It looks like it might be hard to make it to the actual test run section if you have to find a particular failure.

You might be able to do that by separating the "build" step and the "test" steps?

Jammjammjamm (approved):

arscan (approved):
Output looks good!

Screen Shot 2021-10-05 at 2 01 53 PM

Make sure you update this list of tests to include the bulk data tests, when that gets added to the reference server.

PR #52, opened 2021-10-28
Update to support bulk export

hershilpatel:
# Summary
Updated Reference Server to support bulk data operations.  This is a mix of enabling Hapi fhir's bulk data support and adjusting it to pass what inferno program multipatient app tests.

NOTE: There is a bug where sometimes if you run the unit tests again soon after running them (like within a second) one of the tests might fail due to spring transactions.

## New behavior

Inferno Reference Server should now pass the Inferno Program Multi-patient Api Tests.  

## Code changes
Updated hapi version to 5.4
Added Authorization to Bulk operations
Added functionality to include all resources with Patient compartment
Added Bulk Data Interceptor to redirect delete requests

## Testing guidance

Run the unit tests to confirm that they pass.

Make sure you have deleted your existing volume for the reference server if you have one.

The Inferno Reference server should now pass the Multi-Patient Api tests. The params should be:

Bulk Data FHIR URL : http://host.docker.internal:8080/reference-server/r4

Backend Services Token Endpoint: http://host.docker.internal:8080/reference-server/oauth/bulk-token

Bulk Data Client ID: eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6InJlZ2lzdHJhdGlvbi10b2tlbiJ9.eyJqd2tzX3VybCI6Imh0dHA6Ly8xMC4xNS4yNTIuNzMvaW5mZXJuby8ud2VsbC1rbm93bi9qd2tzLmpzb24iLCJhY2Nlc3NUb2tlbnNFeHBpcmVJbiI6MTUsImlhdCI6MTU5NzQxMzE5NX0.q4v4Msc74kN506KTZ0q_minyapJw0gwlT6M_uiL73S4

Bulk Data Scopes: system/\*.\*

Group ID: 64fdf2a5-ebad-4ed0-a512-567970843d49

Patient IDs in exported Group: 85,355

### DIFF
--- a/.github/inferno/docker-compose.yml
+++ b/.github/inferno/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '3'
 services:
   inferno_program:
-    image: onchealthit/inferno-program:release-latest
+    image: onchealthit/inferno-program:latest
     volumes:
       - ./config.yml:/var/www/inferno/config.yml
       - ./wait-for-it/wait-for-it.sh:/var/www/inferno/wait-for-it.sh


arscan:
I'm not sure what kind of data you've loaded on your local instance, but the Group ID on our image with baked-in-data is `64fdf2a5-ebad-4ed0-a512-567970843d49`.  I'm getting errors about Group ID 692 being a DiagnosticReport, and not a group.

arscan:
Could you also please update this page to reflect the added capabilities, and the necessary information regarding what group id to use etc etc?  http://host.docker.internal:8080/reference-server/

hershilpatel:
@arscan I have made the required changes, it should be ready for review (note the different testing data in the description)

radamson (approved):
Looks good overall and all the bulk data tests (aside from the TLS tests since I'm testing locally) pass.

https://user-images.githubusercontent.com/41651655/164513234-bac2a019-f5c3-4da1-bc39-2ee14ed1503f.mov

radamson:
After approving this and poking around some more I noticed that some of the URLs on the reference-server landing page weren't updated. I'll work on a fix.

Screen Shot 2022-04-22 at 8 54 22 AM

arscan:
Do we care that the GroupID is not human-friendly, unlike our Patient IDs?  Its nice to be able to just type in values sometimes, and not have to search for the GroupID because it is a non-memorable UUID.  This is a level-of-polish kind of thing.  Options:

1) Leave as-is -- Go to the reference server's config page to pull out GroupID if you are ever using this outside of a 'preset', or figure out that SAMPLE_TOKEN works to query the /Group resource and grab it with Postman/etc.  Just make sure the preset is right (we will have to change the preset regardless for the URL, so its not like we aren't changing that already)
2) PUT a resource on any server with Group ID 1 manually.  That will not persist so not a good idea.
3) PUT a resource on a running copy of this server with Group ID 1 manually, then export the initdb.sql (I think directions in README), and commit that update.  That is better but if anyone uploads new resources from scratch using the upload.rb script it won't get a human readable ID... this will matter when we get US Core 4, 5 resources generated.
4) Update the upload.rb script to not PUT that UUID group ID (just leave ID blank, or use a POST, so we get a friendlier ID that HAPI gives us), run this server against a fresh psql db instance, export the psql, and add the file to the commit.

Technically, 4 would be ideal to add this extra polish.  Or we can just kick this down the road to when we do US Core v4/5 resources.  Someone other than me on the team needs to be aware of this process though.

arscan:
Do we care that the GroupID is not human-friendly, unlike our Patient IDs?  Its nice to be able to just type in values sometimes, and not have to search for the GroupID because it is a non-memorable UUID.  This is a level-of-polish kind of thing.  Options:

1) Leave as-is -- Go to the reference server's config page to pull out GroupID if you are ever using this outside of a 'preset', or figure out that SAMPLE_TOKEN works to query the /Group resource and grab it with Postman/etc.  Just make sure the preset is right (we will have to change the preset regardless for the URL, so its not like we aren't changing that already)
2) PUT a resource on any server with Group ID 1 manually.  That will not persist so not a good idea.
3) PUT a resource on a running copy of this server with Group ID 1 manually, then export the initdb.sql (I think directions in README), and commit that update.  That is better but if anyone uploads new resources from scratch using the upload.rb script it won't get a human readable ID... this will matter when we get US Core 4, 5 resources generated.
4) Update the upload.rb script to not PUT that UUID group ID (just leave ID blank, or use a POST, so we get a friendlier ID that HAPI gives us), run this server against a fresh psql db instance, export the psql, and add the file to the commit.

Technically, 4 would be ideal to add this extra polish.  Or we can just kick this down the road to when we do US Core v4/5 resources.  Someone other than me on the team needs to be aware of this process though.

radamson:
> Do we care that the GroupID is not human-friendly, unlike our Patient IDs? Its nice to be able to just type in values sometimes, and not have to search for the GroupID because it is a non-memorable UUID. This is a level-of-polish kind of thing. Options:

Will address in a separate PR.

PR #57, opened 2022-04-25
Add Group with id `1a`

radamson:
Adds a Group with id `1a` which includes Patient resources with id `85` and `355`

Updates github action to use this Group.

_Note: HAPI prevents you from clients from setting purely numeric characters (I'm guessing because it might cause with the [default sequential numeric strategy](https://hapifhir.io/hapi-fhir/apidocs/hapi-fhir-storage/ca/uhn/fhir/jpa/api/config/DaoConfig.IdStrategyEnum.html))_

I created FI-1548 to tackle improving the resource loading strategy overall.


arscan (approved):
I'm very glad that we put in the inferno checks here -- we know that the update made into the db properly because of the end-to-end test with group id 1a 👍 

~~Note that we will have to make sure that the initdb.sql update gets into the deployment repo.  Though that does make me wonder: why don't we just bake the data into the image so we don't have to remember to bring this data around everywhere?  I think being targetted in the use of this repo (g)(10) only, but works 100% out of the box, might be a good idea.~~

Nevermind, I forgot that this is setup up to use an external pgsql db, outside of this image.  Though it would be neat if somehow we didn't have to copy this file around by somehow getting the pgsql db to grab it from this image.

PR #58, opened 2022-04-26
FI-1499 Sub Claim

Jack-Fraser16:
# Summary
Added the subject identifier claim to the OpenID token. 

## New behavior
See content of **id_token_payload_json**

Screen Shot 2022-04-26 at 12 04 54 PM

## Code changes
Simple addition of a subject string. 

## Testing guidance
Run core outside of docker with g10 enabled and see the output of test 1.3.01 to see the sub claim in the OpenID token payload.


Jack-Fraser16:
Quick point just to cover the bases - I read a bit about 'public' vs. 'pairwise' sub identifiers. Based off the ticket, it seems that we're just adding/checking for the presence of the sub claim, and are not concerned with whether its public or pairwise. Is this correct?

PR #60, opened 2022-04-27
Inferno Reference Server v1.1.0

radamson:
# Inferno Reference Server v1.1.0

- Add support for Bulk Data Group Export (#52)
- Add additional Group resource with id `1a` (#57)
- Updates cache busting query strings (#59)


Jack-Fraser16 (approved):
Still making my way through [these changes](https://github.com/inferno-framework/inferno-reference-server/pull/52), but that's more so I understand everything going forward.

arscan (approved):

PR #63, opened 2022-08-26
FI-1660: Prevent Writes

Jack-Fraser16:
# Summary
By default, reference server prevents writes of any kind.

## New behavior
Creating/updating resources is no longer supported. Trying to do so will result in an` OperationOutcome` error with a message that specifies the server is read-only and the attempted operation is not supported.

## Code changes
Created and integrated a ReadOnly receptor into the main server class. The receptor is turned on if `READ_ONLY=true`, either in the environment (Docker) or as a CLI flag (run `mvn jetty:run -DREAD_ONLY=true`). Added unit tests.

## Testing guidance
Run the server locally.  Use Postman (or some other http interface), to attempt to write to the local ref server instance. Additionally, run a local instance of G10 (**note**: run G10 using ruby, not docker, to avoid docker's localhost quirks) with the ref server input pointing to the running local ref server instance. Confirm this reference server updated does not change G10 behavior.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '3'
 
 services:
-  db:    
+  db:
     build:
       context: .
       dockerfile: Dockerfile.database
@@ -20,6 +20,7 @@ services:
     environment:
       - POSTGRES_HOST=db
       - CUSTOM_BEARER_TOKEN=SAMPLE_TOKEN
+      - READ_ONLY=true
     networks:
       - fhirnet
     depends_on:
@@ -27,6 +28,6 @@ services:
 
 networks:
   fhirnet:
-  
+
 volumes:
   fhir-pgdata:
--- a/src/config.properties
+++ b/src/config.properties
@@ -1,1 +1,0 @@
-READ_ONLY=false
--- a/src/config.properties
+++ b/src/config.properties
@@ -1,0 +1,1 @@
+READ_ONLY=false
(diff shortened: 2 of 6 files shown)


yunwwang:
I am a little bit concerning about this solution because occasionally we do need to quickly "fix" data on the reference server. Is it possible to use SMART approach so that current testing credentials do not have write access and we can create an admin account which can update resources on the server. Just some thoughts.

Jack-Fraser16:
> I am a little bit concerning about this solution because occasionally we do need to quickly "fix" data on the reference server. Is it possible to use SMART approach so that current testing credentials do not have write access and we can create an admin account which can update resources on the server. Just some thoughts.

I can definitely take that SMART approach. Considering the quick 'fix' use case for the current env var approach, I think what we can do on deployment is bring the container with the reference server down, change the READ_ONLY env variable, stand it back up, update the data, bring it back down, change the variable back to READ_ONLY, then stand it up again. It's clunky and not as clean as what you're suggesting, but including an env var was part of the ticket description.

radamson (approved):
Looks good.

I saw the following as expected when trying to write to the server.

```json
{
	"resourceType": "OperationOutcome",
	"issue": [
		{
			"severity": "error",
			"code": "processing",
			"diagnostics": "Server is currently `read-only`: the CREATE operation is not allowed."
		}
	]
}
```

PR #67, opened 2022-09-15
Update initdb.sql to us-core-4 compatible

yunwwang:
[#]([url](url)) Summary
This ticket updates reference server to be us-core-v4 compatible.

## Code changes
Update initdb.sql, which initializes reference server with us-core-v4 data
Update two patient's JSON file with corresponding changes also. These two JSON files are not loaded during server initialization but could be loaded to external servers.

## Testing guidance
1 Follow the READ.me to re-initialize reference server, or
2. Run US Core Test Kit v4.

Result:
There should be two errors only. One on CapabilityStatement, which will be addressed in a different ticket. One on the DocumentReference due to tx.fhir.org could 
not validate new DocumentReference code, which is ok. 

Also attach zip file for FHIR transaction Bundle which updates data in existing reference server. 
[us-core-server-400-update.json.zip](https://github.com/inferno-framework/inferno-reference-server/files/9579170/us-core-server-400-update.json.zip)


arscan:
I don't really care that much, but we really should only have stuff loaded in there that matter.  The CapabilityStatement doesn't affect the behavior of the server at all by uploading it, so we simply shouldn't have it on there (we certainly don't have /CapabilityStatement read/searches in our tests).  Same with StructureDefinitions, if those are in there too.

us-core-r4-resources/us-core-server-400-update.json
```diff
@@ -0,0 +1,11213 @@
+{
```
arscan, line 1: While this is a nice file to have, I don't think it should be part of the repo because as soon as we merge this in, it is basically obsolete, right?  Should this file just be an attachment to the PR or something like that?

In other words, if you started from 'scratch' after this is merged in, what would this file do?  You'd have no way of initializing the server in the "old" config, because you've updated the initdb.sql.  I could see this file just being confusing after we get through this process.

(an alternate approach could be to have 'migrations', where you always leave the original transaction and then just keep adding update transactions, but that  won't work generically as-is because these are 'hardcoded' with IDs that only happen to exist in our snapshot, not when others re-run the old transactions)

arscan: In short, I'd put this file as an attachment to the PR, and remove the file from the actual repo.

yunwwang: Removed

Jammjammjamm:
The data seems good to me for v3 and v4 in the g10 and us core test kits.

arscan (approved):
Looks good, got the g10 tests in both single and multi apis to pass for v3 and v4 of us core 🥳 

Please make sure this update gets reflected in a PR to https://github.com/onc-healthit/inferno-deployment

And that this PR eventually gets in too https://github.com/inferno-framework/uscore-data-sets/pull/5

PR #70, opened 2022-10-05
FI-1672: Remove patient-picker from EHR launch

Jack-Fraser16:
# Summary
Removed the patient-picker page from the EHR Launch flow. 

## New behavior
Upon receiving an authorization request from within an EHR launch – as denoted by the url containing the `launch` parameter – the reference server prevents navigation to the patient-picker and allows for immediate scope granting. 

## Code changes
Altered the `authorization.js` file to check the url for the `launch` parameter and to append `patient_id=launch` to the url, if the launch parameter has an expected value. By adding a `patient_id` parameter, the patient-picker is circumvented.

## Testing guidance
Run the reference server locally and use it as the FHIR endpoint when running the g10 suite locally (I recommend using the command `G10_VALIDATOR_URL=http://localhost:4545 INFERNO_HOST=http://localhost:4567 ASYNC_JOBS=false bundle exec puma` to avoid docker/localhost quirks). Run the EHR launch group and follow the normal flow.

### DIFF
--- a/src/main/webapp/WEB-INF/templates/app-launch.html
+++ b/src/main/webapp/WEB-INF/templates/app-launch.html
@@ -6,37 +6,50 @@
 
     <title>App Launch</title>
     <meta content="Authorize Cindehr" name="description">
-    
-    <link href="js/lib/bootstrap-4.5.3-dist/css/bootstrap.min.css" rel="stylesheet"> 
+
+    <link href="js/lib/bootstrap-4.5.3-dist/css/bootstrap.min.css" rel="stylesheet">
     <link href="css/site.css" rel="stylesheet">
 </head>
 
 <body>
 
-	<div th:replace="tmpl-banner :: banner"></div>
-	
-	<br>
-	<br>
-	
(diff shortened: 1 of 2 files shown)


Jammjammjamm:
I get this when I do the redirect to the auth endpoint:

Screen Shot 2022-10-14 at 9 29 29 AM

arscan (approved):
I was able to get this to work fine for all SMART v1 tests.

However, it broke on SMART v2, presumably because this branch hadn't been rebased against main which has the POST fix.  Could you please confirm that this works fine for SMART v2 once it is merged in?

Jack-Fraser16:
mine's also working fine, but the authorization url on mine is different than @Jammjammjamm  -- it has a patient_id query parameter on the end (which is the expected behavior). It routes to the authorization page, checks whether the scope parameter is there/includes a valid patient id, and then appends it to the url/redirects. However, I realized the redirect/adding to the url is unneccessary based off how things are changed in Encounter PR, so I don't think its worth digging into why yours doesnt append since it won't work that way soon.

PR #72, opened 2022-10-14
FI-1758: Address "Invalid Token" Error raised by getAllEncounters

Jack-Fraser16:
# Summary
Any method in the `FhirUtils` class (such as `getAllEncounters`) that receives a Bundle and then extracts every resource in that Bundle into an array will issue follow-up `GET_PAGE` request(s), if the Bundle is big enough to need muliple pages. However, the Bearer Token was not being passed as part of those requests, causing them to be denied and the "Invalid Token" error to be thrown. This was actually not an error unique to `getAllEncounters`, but every `getAll` method in `FhirUtils`. However, it only arose while testing `getAllEncounters` because Encounters are the only resource whose quantity requires multiple pages in the returned Bundle.

## New behavior
No longer throws `Invalid Token` when calling `getAllEncounter`.

## Code changes
Add Bearer Token to subsequent `GET_PAGE` request when accessing pages in Bundle.

## Testing guidance
This isn't easy to test because the `FhirUtils.getAllEncounters` method is not directly used in any part of the API, and creating a unit test would mean populating the server with enough encounters to require multiple pages in the bundle (time consuming and repetitive to write). My approach was to add a basic mapping that just calls the method and then hit it with the server running to make sure no error is raised and the Encounters are collected. If testing using Docker, make sure to change the `READ_ONLY` flag, as the GET_PAGE request will be blocked by the read-only interceptor (there's a separate, incoming PR to address that).

### DIFF
--- a/src/main/java/org/mitre/fhir/app/launch/AppLaunchController.java
+++ b/src/main/java/org/mitre/fhir/app/launch/AppLaunchController.java
@@ -20,9 +20,7 @@ public class AppLaunchController {
   @GetMapping(path = "/fhir-server-path")
   public ResponseEntity<String> getFhirServerPath(HttpServletRequest request) {
     String fhirServerBaseUrl = FhirReferenceServerUtils.getFhirServerBaseUrl(request);
-    ResponseEntity<String> responseEntity = new ResponseEntity<String>(fhirServerBaseUrl,
-        HttpStatus.OK);
-    return responseEntity;
+    return new ResponseEntity<String>(fhirServerBaseUrl, HttpStatus.OK);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


arscan (approved):
I'll go ahead and approve this, but FHIRUtils + unit tests + the `https://inferno.healthit.gov/reference-server/oauth/authorizeClientId/SAMPLE_CONFIDENTIAL_CLIENT_ID` route is something like 250 lines of code to simply get a list of patients (and I suppose now encounters too).  That's a lot of code to maintain that barely gives you anything more than the standard FHIR API that we get for free with HAPI.

Because every line of code requires maintenance, I'd think the better approach for this type of demo server is to just have the JS front-end use the standard FHIR API with the 'universal token' SAMPLE_TOKEN (do a GET /Patient, GET /Encounter?patient=[patient]... maybe with _sort or _count or a _include if you are feeling fancy).  And its ok to not let users choose every encounter... just having a handful of choices is good enough.

Or if we want to make this a bit more realistic by not having this UI make a call to the 'protected' fhir server that can be bypassed with a universal token, just render the page with the patient ids and encounter ids in them... and do that with the ~10 lines of code needed without pagination and scrap the rest of this.

PR #73, opened 2022-10-14
FI-1759: Allow GET_PAGE requests in ReadOnly Interceptor

Jack-Fraser16:
# Summary
Allow GET_PAGE requests to be accepted when the server is in ReadOnly mode. 

## New behavior
GET_PAGE requests will not be blocked nor return a 'the server is currently read only' response/error, but will be accepted by the server.

## Code changes
Added the GET_PAGE request to the list of acceptable requests within the interceptor.

## Testing guidance
Do a search and attempt to read second page of results.

### DIFF
--- a/src/main/java/org/mitre/fhir/ReadOnlyInterceptor.java
+++ b/src/main/java/org/mitre/fhir/ReadOnlyInterceptor.java
@@ -27,7 +27,8 @@ public class ReadOnlyInterceptor {
             && theOperation != RestOperationTypeEnum.VALIDATE
             && theOperation != RestOperationTypeEnum.VREAD
             && theOperation != RestOperationTypeEnum.GET_PAGE
-            && theOperation != RestOperationTypeEnum.EXTENDED_OPERATION_INSTANCE) {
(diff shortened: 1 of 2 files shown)


Jammjammjamm:
More operations need to be added for bulk data:

Screen Shot 2022-10-19 at 11 03 52 AM

arscan:
It paged results fine but I think we should consider any similarly blocked requests like bulk data to be in scope here, so please fix the one that Steve found @Jack-Fraser16.

Jack-Fraser16:
Ran against US Core v3.1.1 / SMART v1 / Bulk Data v1 and Bulk Data v2 and only saw the expected TLS and validation errors. Seems these are the necessary allowed actions.

arscan (approved):
Ran it against single patient api and multi-patient api and this removed the problems that I saw prior to switching to this branch.

I do think we should consider having the default mode be read-only though, seems safer.  We probably should inform the user on the 'landing page' if it has write-enabled too... in my case i'd prefer to not have to issue a request to verify which config it is in.  Maybe other tickets?

PR #74, opened 2022-10-19
FI-1727: Encounter Context in EHR Launch

Jack-Fraser16:
# Summary
Add support for selecting the encounter from which an EHR launch occurs, and then passing that encounter information throughout the authorization flow.

## New behavior
The Reference Server App Launch page now includes a dropdown containing the encounters associated with each stored patient. Both the selected patient and encounter are passed to app via the launch context parameter. 

Screen Shot 2022-10-19 at 12 16 03 PM

The patient-picker is then ignored (will include changes from this [PR](https://github.com/inferno-framework/inferno-reference-server/pull/70)). The encounter selected is then added to the token (and the refresh token). In this example, the app launch occurred from within Encounter 301 of Patient 232.

Screen Shot 2022-10-19 at 12 29 32 PM

## Code changes

- The Patient ID dropdown was hard-coded, but is now dynamic (as is encounter). These dropdowns take a half second to populate, which is not ideal – I have an idea of how to make it less obvious, but wanted to get feedback first.
- Create 'ehr-launch-context-options' – and created/modified related utils – to get the data for populating the Patient ID and Encounter ID dropdowns.
- Added JavaScript on top of the Reference Server App Launch html page and added dropdowns to that page.
- Altered the authorization endpoint JavaScript to try and extract the patient and encounter IDs from the launch parameter, if possible, and adjust the flow accordingly (don't go to `patient-picker` and add the ids to the authorization code)
- `generateBearerToken` takes encounterId as an argument and includes it within the token, if it's provided. Otherwise, it gets the first encounter by searching on the the given patientId.
- Added encounterId to token. 
- Various unit tests.
- General clean-up as suggested by `checkstyle` and IntelliJ (anything not mentioned above was clean-up).

## Testing guidance
Run without docker (make sure to change the DB config first, as specified in the README) and also run G10 locally using `G10_VALIDATOR_URL=http://localhost:4545 INFERNO_HOST=http://localhost:4567 ASYNC_JOBS=false bundle exec puma`. Run SMART v1 and confirm that the selected encounter is added as part of the token in EHR launch, and that the Standalone and Limited Launches are unchanged.

### DIFF
--- a/src/test/java/org/mitre/fhir/utils/TestFhirReferenceServerUtils.java
+++ b/src/test/java/org/mitre/fhir/utils/TestFhirReferenceServerUtils.java
@@ -11,23 +11,21 @@ import org.springframework.mock.web.MockHttpServletRequest;
 
 public class TestFhirReferenceServerUtils {
 
-  private static final MockHttpServletRequest mockRequest = new MockHttpServletRequest();
-
   @Test
   public void testGetServerBaseUrl() {
-    String baseUrl = FhirReferenceServerUtils.getServerBaseUrl(mockRequest);
-    Assert.assertEquals("http://www.example.org:123/reference-server", baseUrl);
+    String baseUrl = FhirReferenceServerUtils.getServerBaseUrl(getMockRequest());
+    Assert.assertEquals("http://www.example.org/reference-server", baseUrl);
   }
 
   @Test
   public void testGetServerBaseUrlWithHttpDefaultPort() {
-    mockRequest.setServerPort(80);
-    String baseUrl = FhirReferenceServerUtils.getServerBaseUrl(mockRequest);
+    String baseUrl = FhirReferenceServerUtils.getServerBaseUrl(getMockRequest());
     Assert.assertEquals("http://www.example.org/reference-server", baseUrl);
   }
 
   @Test
   public void testGetServerBaseUrlWithHttpsDefaultPort() {
+    MockHttpServletRequest mockRequest = getMockRequest();
     mockRequest.setScheme("https");
     mockRequest.setServerPort(443);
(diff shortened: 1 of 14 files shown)


src/test/java/org/mitre/fhir/utils/TestFhirReferenceServerUtils.java
```diff
@@ -2,62 +2,42 @@
...
+
+import ca.uhn.fhir.rest.client.api.IGenericClient;
 import org.junit.Assert;
+import org.junit.BeforeClass;
 import org.junit.Test;
 import org.springframework.mock.web.MockHttpServletRequest;
 
 public class TestFhirReferenceServerUtils {
 
+  private static final MockHttpServletRequest mockRequest = new MockHttpServletRequest();
```
arscan, line 14: do these tests always run in the same order?  My understanding is that generally it is a bad idea to assume test frameworks will run from top to bottom, and it seems like you might get different results if they are run in different orders.

arscan (approved):
I made some changes:

* added cache busting for the links to js files so the new ones come up in prod and won't serve the old js
* improved language when patient picker does not yet have data which has always bugged me
* added explanatory language to the app launch page on the top
* added a default 'loading' option on the app launch page for patients and encounters because an empty select box with no options just looks uprofessional
* removed a seemingly unneeded redirect that i couldn't understand the purpose of on the authorization page.

I tested just about everything out for SMART v1 and SMART v2 for US Core, and did the relevant ones for US Core v5 (e.g. encounter tests now pass because they get the right field).  We definitely will want to do an end-to-end test in dev for everything after all these updates though.

PR #75, opened 2022-10-25
Cleanup reference server landing page

chris-at-work:
# Summary

* Change copy/content from ticket
* Whitespace changes

## New behavior

None.  Just copy / content changes.

## Code changes

None except markup.

## Testing guidance

Boot app and look at HTML / CSS.  I was unable to do this because of a maven timeout.

### DIFF
--- a/src/main/webapp/WEB-INF/templates/landing.html
+++ b/src/main/webapp/WEB-INF/templates/landing.html
@@ -1,28 +1,29 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="utf-8">
 
-  <title>Inferno Reference Server</title>
-  <meta name="description" content="Inferno Reference Server">
-  
-  <link href="/reference-server/app/js/lib/bootstrap-4.5.3-dist/css/bootstrap.min.css" rel="stylesheet"> 
-  <link href="/reference-server/app/css/site.css" rel="stylesheet">
+<head>
+	<meta charset="utf-8">
+
+	<title>Inferno Standardized API Demonstration Server</title>
+	<meta name="description" content="Inferno Reference Server">
+
+	<link href="/reference-server/app/js/lib/bootstrap-4.5.3-dist/css/bootstrap.min.css" rel="stylesheet">
+	<link href="/reference-server/app/css/site.css" rel="stylesheet">
 </head>
 
 <body>
 	<div th:replace="tmpl-banner :: banner"></div>
-	
-	<br>
-	<br>
-	
-    <div id="pageContent" class="container">
-    
-    
-   		
-   		<p>This server is designed to meet the requirements of the Inferno Program Certification Tests.  This includes:</p>
 
-<p>Support for SMART-on-FHIR, including:</p>
+	<br>
+	<br>
+
+	<div id="pageContent" class="container">
+
+		<p>This server demonstrates a mock API that is capable of passing the automated Inferno ONC Certification (g)(10)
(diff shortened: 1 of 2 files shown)


src/main/webapp/WEB-INF/templates/tmpl-banner.html
```diff
@@ -1,10 +1,9 @@
-<!DOCTYPE html>
+<!doctype html>
 <html lang="en">
 <div th:fragment="banner">
     <div class="row navbar navbar-light inferno">
         <div class="col-12 text-center">
-        	
-            <h4 class="">Inferno Demo FHIR Server</h4>
+            <h4 class="">Inferno Reference Server</h4>
```
Jack-Fraser16, line 6: Should the title be "Inferno ONC Standardized API Demo Server"? I see that it is the subheading, but I think it belongs at the top. Unlike the test kits which exist to be used outside the context of G10, the only real use case for this server is within G10, so I think it's fine to just label it this way.

src/main/webapp/WEB-INF/templates/landing.html
```diff
@@ -1,127 +1,144 @@
...
-		      <input type="text" readonly class="form-control-plaintext" id="sampleGroupId" th:value="${groupId}">	    	
-		    </div>
-		  </div>
-		  
-		  
+			<h3>Inferno Standardized API Demonstration Server</h3>
+			<p>
+				<a href="https://inferno.healthit.gov/reference-server/r4">This server</a> is protected by a simple mock
+				OAuth2.0 implementation,
+				which can be bypassed by using SAMPLE_TOKEN bearer token. Use patient <code>85</code> or <code>355</code>.
```
Jack-Fraser16, line 111: I would move this up to reside under the blurb you added at the top of the page, maybe with a line of spacing between them, but no 'Inferno Standardized API Demonstration Server' subheading (since that should be the title).

Jack-Fraser16 (approved):
LGTM - I think @arscan should take a quick look before merge just to make sure the wording is on point and nothing is missing.

src/main/webapp/WEB-INF/templates/landing.html
chris-at-work: I changed the title to be `Inferno Standardized API Demonstration Server` and moved the paragraph up.  Hard to see since I can't preview because of Java problems.

arscan (approved):
Looks good.

We will need to make sure that this is updated when we bring in new versions of the standards soon.

Also, I noticed that there is a separate token endpoint for backend services than for smart app lauch. While our tests pass this because there are light requirements around discovery requirements for backend services, in theory `.well-known/smart-configuration` should point to a single token endpoint that supports both uses of OAuth.  So, in other words, we need to unify the SMART App Launch token endpoint here with the BackendServices endpoint.  Lets open a ticket for this.

Jack-Fraser16:
> Looks good.
> 
> We will need to make sure that this is updated when we bring in new versions of the standards soon.
> 
> Also, I noticed that there is a separate token endpoint for backend services than for smart app lauch. While our tests pass this because there are light requirements around discovery for backend services in g10, in theory `.well-known/smart-configuration` should point to a single token endpoint that supports both uses of OAuth. So, in other words, we need to unify the SMART App Launch token endpoint here with the BackendServices endpoint. Lets open a ticket for this.

Made two tickets -- one for updating as new versions of the IGs are supported and one for consolidating to a single token endpoint.

PR #79, opened 2022-12-05
FI-1726: PKCE support

Jammjammjamm:
This branch adds PKCE support. It refactors the code generation so that instead of being a series base64-encoded strings with period delimiters, the code is now just a base64-encoded json object so that it's easier to add more fields as needed.

### DIFF
--- a/src/main/java/org/mitre/fhir/utils/FhirReferenceServerUtils.java
+++ b/src/main/java/org/mitre/fhir/utils/FhirReferenceServerUtils.java
@@ -77,6 +77,28 @@ public class FhirReferenceServerUtils {
   }
 
   /**
+   * Create a code embed with the actual code, scopes and patient id.
+   *
+   * @param actualCode the code itself for auth purposes
+   * @param scopes the scopes the user selected for this token
+   * @param patientId the selected patientId
+   * @param codeChallengeMethod PKCE
+   * @param codeChallenge PKCE
+   * @return string representation of code containing the input code, scopes, and patientId
+   */
+  public static String createCode(String actualCode, String scopes, String patientId, String codeChallengeMethod, String codeChallenge) {
+    JSONObject code = new JSONObject();
+
+    if (actualCode != null) code.put("code", actualCode);
+    if (scopes != null) code.put("scopes", scopes);
+    if (patientId != null) code.put("patientId", patientId);
+    if (codeChallengeMethod != null) code.put("codeChallengeMethod", codeChallengeMethod );
+    if (codeChallenge != null) code.put("codeChallenge", codeChallenge);
+
+    return Base64.getEncoder().encodeToString(code.toString().getBytes());
+  }
+
+  /**
    * Create the Authorization Header value.
    * 
    * @param accessToken the access token value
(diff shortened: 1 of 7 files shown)


src/main/webapp/js/authorize.js
```diff
@@ -2,142 +2,148 @@ window.mitre = window.mitre || {};
...
-            // base64 encoding that is escaped so it can be used in a url
-            let base64URLEncodedScopes = btoa(selectedScopes);
+      // get checked scopes
+      let selectedScopes = "";
+      $('#scopes [name="scopeCheckbox"]:checked').each(function(index, checkbox) {
+        selectedScopes += `${checkbox.value} `;
+      });
 
-            let code = sampleCode + "." + base64URLEncodedScopes + "." + base64URLEncodedPatientId;
+      const codeParams = Object.fromEntries(Object.entries({
```
Jammjammjamm, line 127: This file is mostly just indentation changes. The real change is this modification to how we construct the code.

arscan: So for 3-legged oauth, we have javascript create the code, and the java backend will parse it, right?  Creating the code in java is only for backend services (2-legged oauth), right?

Jammjammjamm: Backend services doesn't use `code` at all.

arscan: Sorry, misinterpretted why java is ever generating the hashed code.  createCode in java is only for tests, right?

Jammjammjamm: In `FhirReferenceServerUtils`? Yes.

src/main/java/org/mitre/fhir/utils/FhirReferenceServerUtils.java
```diff
@@ -66,10 +67,57 @@ public static String getSmartStyleUrl(HttpServletRequest request) {
...
+   * Create a code embed with the actual code, scopes and patient id.
+   *
+   * @param actualCode the code itself for auth purposes
+   * @param scopes the scopes the user selected for this token
+   * @param patientId the selected patientId
+   * @param codeChallengeMethod PKCE
+   * @param codeChallenge PKCE
+   * @return string representation of code containing the input code, scopes, and patientId
+   */
+  public static String createCode(
```
arscan, line 95: i'm having trouble identifying where this 5 argument variant is called?  I see multiple 3 argument variants called in our unit tests, but can't find where this one is called.  Did we mean to add a test for pkce code generation but didn't?  We probably should get rid of the first and just use this one (and pass 2 null params at the end if pkce not enabled).

Jammjammjamm: Look at the bottom of `TestAuthorization.java`. I could've made it a single method, but I didn't want to have to go through and change every invocation of `createCode`. I think it makes a reasonable amount of sense to have a version without PKCE stuff.

arscan: Sorry, i just did a search in the github webui but that diff was too big to show so i didn't search in there by mistake :)

PR #81, opened 2022-12-08
FI-1771: Update landing page to mention new standards supported

Jammjammjamm:
Updated the landing page:

Screen Shot 2022-12-08 at 9 50 58 AM

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
 
   <groupId>org.mitre.fhir</groupId>
   <artifactId>inferno-reference-server</artifactId>
-  <version>1.1.1</version>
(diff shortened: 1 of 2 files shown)


arscan:
I think the request may have been against the README (not sure, but makes sense to describe what the specific standards are there).  Should we add more specificity in there now or create a new ticket?

Jammjammjamm:
I'd like that to be a separate ticket. I want to do a reference server release ASAP so we can get it up on dev.

Jammjammjamm:
I also just bumped the version so that I don't need to do another PR to release this.

arscan (approved):

PR #82, opened 2022-12-15
FI-1833: Bulk Data Fixes

Jammjammjamm:
This branch updates the bulk data implementation to support SMART v2 scopes and export the additional resources needed by US Core 5.


arscan (approved):
Along with the new update to the tests, all the bulk data tests now pass with this for us core 5.

Note: I was having trouble with stale exports, so i ended up wiping the volume (also forgot to update the authorization endpoint to my local running instance)

PR #86, opened 2023-01-03
FI-1740: Handle deleted bulk requests

Jammjammjamm:
This branch:
- Removes caching from bulk data, so all requests won't share the same id
- Makes the server return a 404 for status requests of cancelled jobs

To test:
- Run the g10 bulk data tests
- Get the status url and bearer token from the delete request
- Make a status request using that url and token in postman. You should get a 404


arscan (approved):
Followed the testing instructions and it worked as advertised.

As we had discussed, we'll want to keep an eye on resource usage to ensure that removing caching doesn't use up too much disk space, but it sounds like HAPI should handle this automatically.

PR #88, opened 2023-01-26
FI-1871: Add missing profiles to capability statement

Jammjammjamm:


arscan (commented):
While we are updating this, could you scrub this for any capabilitystatement-expectation extensions (they may no sense for capabilitystatement 'kind' = instance)?  And also update the language so that it states that it supports more than just US Core v3.1.1?

This line should also be updated?
```
<implementationGuide value="http://hl7.org/fhir/us/core/ImplementationGuide/hl7.fhir.us.core|3.1.1"/>
```

Or if you want I can make new tickets for this.

arscan (approved):

PR #89, opened 2023-01-31
FI-1770: Default to read only

Jammjammjamm:
# Summary
Server now defaults to read only mode if the `READ_ONLY` environment variable is not set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,13 +65,14 @@ To revoke a token, go to `reference-server/oauth/token/revoke-token`
 
 The tokens are currently saved in memory, so if the reference server is restarted, all existing tokens will be invalid
 
-## Running Read-Only Mode
+## Running Read-Write Mode
 
-By default, the Dockerized server runs in read-only mode – meaning, operations modifying the state of the server 
-are not supported. For example, requests to CREATE, UPDATE, or DELETE a resource will receive a `405 Method Not Allowed` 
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -65,13 +65,14 @@ To revoke a token, go to `reference-server/oauth/token/revoke-token`
...
-`./docker-compose.yml`. If running without Docker, the default mode is _not_ read-only. To activate read-only, run 
-`mvn jetty:run -DREAD_ONLY=true` when starting the server.
+## Running Read-Write Mode
+
+By default, the Dockerized server runs in read-only mode – meaning, operations
+modifying the state of the server are not supported. For example, requests to
+CREATE, UPDATE, or DELETE a resource will receive a `405 Method Not Allowed`
+error. To adjust this while running Docker, change the `READ_ONLY` environment
+variable to `false` in `./docker-compose.yml`. If running without Docker, , run
+`mvn jetty:run -DREAD_ONLY=false` when starting the server.
```
arscan, line 75:
```suggestion
variable to `false` in `./docker-compose.yml`. If running without Docker, run
`mvn jetty:run -DREAD_ONLY=false` when starting the server.
```

arscan (approved):
Works as intended... I tried no env variable, env variable set to TRUE, env variable set to false.

I do think this config should be visible on the reference server landing page so a misconfiguration would be easy to spot, but that is outside the scope of this ticket.  I created a new ticket internally (low priority).

PR #96, opened 2023-05-12
FI-1969: Unify token endpoints

Jammjammjamm:
# Summary
The server's advertised token endpoint is `oauth/token`, but bulk data used a separate unadvertised `oauth/bulk-token` endpoint. The server advertised supporting client credentials for auth, but that was only supported at the bulk token endpoint, not the normal token endpoint which was advertised. This PR eliminates the `oauth/bulk-token` endpoint so tokens from bulk data can be obtained from the normal token endpoint.

## New behavior
The `oauth/bulk-token` endpoint has been removed, and bulk data now needs to use `oauth/token` as its token endpoint.

## Testing guidance
```
docker-compose build
docker-compose up
```
Then run the g10 tests (in ruby), changing the host to `http://localhost:8080`, and the bulk token endpoint to end in `/token` rather than `bulk-token`. All tests should pass except for TLS tests.


rpassas (approved):
Only change I made was to have changes reflected in the landing page / yaml in the server. Changes made work as expected - new server works with /token not /bulk-token (opposite of main branch)

PR #110, opened 2023-10-09
FI-2195: Bump dependencies

dehall:
# Summary
This PR bumps as many dependencies as we can to address security alerts, without bumping the HAPI JPA server from 5.x to 6.x (since that upgrade would probably require significant refactoring/rewriting)

The latest HAPI 5.x release is 5.7.9 so that's what we'll use. I attempted to mix & match JPA server 5.7.9 plus other HAPI libraries version 6.8.3 and it initially appeared to work, but trying to run it resulted in NoClassDefFoundError and the like which may or may not be solvable, so I think it makes the most sense to keep all HAPI libraries at the same version.

This PR also bumps a few other libraries and nested dependencies via the `<dependencyManagement>` section of `pom.xml`. Marked as Draft as this currently addresses the "critical" level alertd but I also want to look at "high" level alerts

## New behavior
Should be none. Any observed differences are unintentional and should be reviewed.

## Code changes
Outside of `pom.xml` changes are only to support any API changes in our dependencies.

## Testing guidance
In addition to just testing the various features to make sure nothing broke, you can run `mvn org.owasp:dependency-check-maven:check` to get a report of vulnerabilities in dependencies (since the dependabot alerts only work on main)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -12,10 +12,10 @@
   <properties>
     <maven.compiler.source>11</maven.compiler.source>
     <maven.compiler.target>11</maven.compiler.target>
-    <hapi.version>5.4.0</hapi.version>
+    <hapi.version>5.7.9</hapi.version>
     <jetty_version>9.4.35.v20201120</jetty_version>
-    <checkstyle.plugin.version>3.1.1</checkstyle.plugin.version>
-    <checkstyle.version>8.32</checkstyle.version>
+    <checkstyle.plugin.version>3.3.0</checkstyle.plugin.version>
+    <checkstyle.version>10.12.4</checkstyle.version>
     <checkstyle.config>config/checkstyle/checkstyle.xml</checkstyle.config>
     <sevntu.checkstyle.plugin.version>1.37.1</sevntu.checkstyle.plugin.version>
   </properties>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -223,6 +231,133 @@
...
   </dependencies>
 
+  <!--
+    dependencyManagement allows us "to directly specify the versions of artifacts
+    to be used when they are encountered in transitive dependencies
+    or in dependencies where no version has been specified".
+    For example, if a direct dependency above brings in a nested dependency
+    with a vulnerability, we can bump the nested dependency here.
+    (Assuming the new version is still compatible, etc)
+    This section should be reviewed every time we bump direct dependencies.
```
Jammjammjamm, line 241: So whenever we bump a direct dependency, we need to check all of its dependencies against this list to make sure that we've not specifying a lower version?

dehall: I think the minimum we need to do when we update the direct dependencies is 1) Make sure everything still works, and 2) review the vulnerabilities report. If we do specify a lower version here, either it breaks something, it's got a vulnerability, or it works fine and we don't need to worry about it.
Really my intent with that "This section should be reviewed every time..." is to try to keep this dependency management section as small as possible, so ideally we would check and remove any entries that are no longer needed, but it's not a big deal to leave them in.

Jammjammjamm (approved):

PR #112, opened 2023-10-19
Initial token introspection with incomplete fields.

arscan:
WIP DO NOT MERGE.  Intended to be used right now to support token introspection test development, and will need a few updates prior to merging.

# Summary
Adds token introspection capabilities per openid token introspection standard

## Testing guidance
This is intended to be run against the new token introspection tests under development in https://github.com/inferno-framework/smart-app-launch-test-kit/

There are a few things that are not yet complete, but i figured I'd let the tests under development surface those.

To use, I recommend:

* cloning this repository
* running `docker-compose up --build`.  Note, the first time, the db may be slow to launch because it is loading data, and you'll see an issue in the console and the site won't load. Kill the container and rerun `docker-compose up`, as the faster start to the db after it has been initialized should let the web server load
* confirm that http://localhost:8080/reference-server works and that there is a bullet stating support for 'Token Introspection'
* visit http://localhost:8080/reference-server/r4/.well-known/smart-configuration and there should be a URL for introspection_endpoint' (http://localhost:8080/reference-server/oauth/token/introspect)
*optiona: POST to this endpoint with no auth headers (I haven't added authorization here), and a body of `token=ABC`, and it should return json to the effect of `{'active': false}`

In the smart app launch test kit, assuming a 'ruby development setup', do the following:
* remove from `.env` and `.env.development` the `COMPOSE_PROFILES=keycloak` line -- i was having a little bit of trouble with this interferring with the reference server when running it at the same time, so lets just not start keycloak any more.
* `inferno services start` to start the background services (db, redis, etc)
* `inferno start` to start the application
* visit http://localhost:4567
* select STU2 of the smart app launch
* select 'Inferno Reference Server' preset
* Run 'standalone launch' tests
* change endpoint to http://localhost:8080/reference-server/r4
* click through and the tests should pass, except for TLS
* In test 1.1.01 (discovery), take a look at the output tab and grab the token introspection url from there (will be http://localhost:8080/reference-server/oauth/token/introspect)
* run the token introspection test
* appropriately split up the URL so that the current inputs work (note how we will want a single introspection url input insteead of having them split up in two inputs later)
* turn off 'include scope' thing for keycloak on the bottom
* run the group tests
* You'll get blocked by the fact that the token is not a JWK
* Remove a bunch of code in that test so that it really just verifying that the POST returns 'active = true'
* restart the inferno tests, it should now pass

This should allow us to much quicker iterate on the redesigned smart introspection tests.  Please let me know ASAP if any of these steps are giving you issue, running several docker-compose apps can be a little fiddly.  we might consider adding the reference server to the smart app launch docker-compose config as well as an optional thing so it is part of the 'test kit', but unfortunately that involves copying around a sql init file which is not good, so maybe not.


src/main/java/org/mitre/fhir/authorization/token/TokenManager.java
```diff
@@ -24,8 +31,11 @@ private TokenManager() {
     String customBearerTokenString = System.getenv().get(CUSTOM_BEARER_TOKEN_ENV_KEY);
 
     if (customBearerTokenString != null) {
+      Long expiresIn = 10368000L; // 4 months
       Token customBearerToken = new Token(customBearerTokenString,
           FhirReferenceServerUtils.getScopesListByScopeString(CUSTOM_BEARER_TOKEN_SCOPE_STRING));
+      customBearerToken.setClientId("SAMPLE_CLIENT_ID");
+      customBearerToken.setExp(java.time.Instant.now().getEpochSecond() + expiresIn);
```
dehall, line 38: Not sure if this is really better but you could leverage the java time libraries to do the time math:
```java
      customBearerToken.setExp(Instant.now().plus(4, ChronoUnit.MONTHS).getEpochSecond());
```

src/main/java/org/mitre/fhir/authorization/token/TokenController.java
```diff
@@ -32,4 +36,62 @@ public void revoke(@RequestParam(name = "token", required = true) String tokenSt
...
+
+          String fhirUserUrl = FhirReferenceServerUtils
+              .getFhirServerBaseUrl(request) + "/Patient/" + token.getPatientId();
+          tokenResponse.put("fhirUser", fhirUserUrl);
+          tokenResponse.put("iss", FhirReferenceServerUtils.getFhirServerBaseUrl(request));
+          tokenResponse.put("sub", TokenManager.SUB_STRING);
+        }
+
+      } catch (TokenNotFoundException tokenNotFoundException) {
+        // This doesn't feel quit right, but am staying consistent
```
dehall, line 90: Reading up a little on the token introspection endpoint, this seems fine to me. I'm not sure even logging the exception is that necessary.

dehall (commented):
Looked through the code and I don't have any real concerns, just a couple trivial notes. I want to run through the g10 test kit with this myself tomorrow though

dehall (approved):

PR #135, opened 2024-01-19
FI-2367 Add asserted-date

360dgries:
asserted-date structureDefinition and searchParameter are now included in the resources used to build the server.

Testing process:
Spin up the server and try the following GET, it should yield one result.

`http://localhost:8080/reference-server/r4/Condition?asserted-date=2016-08-10`

### DIFF
--- a/resources/uscore_bundle_asserted-date.json
+++ b/resources/uscore_bundle_asserted-date.json
@@ -242,123 +242,127 @@
       },
       {
           "resource":{
-"resourceType" : "SearchParameter",
-"id" : "us-core-condition-asserted-date",
-"text" : {
-  "status" : "generated",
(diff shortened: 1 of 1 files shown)


yunwwang:
Test passed. Just one question. Does the sequence of loading resources matter? Or say, does the search parameters bundle has to be loaded before patient bundles? If so, should we make this clear either in the README or change the file name to indicate that "this bundle must be load before patient bundles", and "this bundle must be loaded after patient bundles".

360dgries:
> Does the sequence of loading resources matter? Or say, does the search parameters bundle has to be loaded before patient bundles?

I don't think the order should matter.  @dehall can you confirm?

dehall:
I agree the order shouldn't matter but I'm not sure that's 100% guaranteed. From the debugging I did earlier to get this search param working, I saw both combinations in my debugging: an instance where a matching resource was added before the search param (the mother's maiden name one), and an instance where a matching resource was added after (this asserted-date), and both worked.

360dgries:
@yunwwang let me know if you want me to try forcing the server to boot up in different orders and test that it works to fully confirm.

yunwwang:
> @yunwwang let me know if you want me to try forcing the server to boot up in different orders and test that it works to fully confirm.

No problem. I think this is fine for now. We will decide when there is another similar request.

yunwwang (approved):

PR #139, opened 2024-03-15
FI-2424: Support for SMART asymmetric confidential client auth

dehall:
# Summary
This PR adds support for the SMART Asymmetric Confidential client authentication method: https://build.fhir.org/ig/HL7/smart-app-launch/client-confidential-asymmetric.html
Both `EC384` and `RS384` methods are supported since those are the 2 the SMART App Launch test supports.

## Code changes
 - The main code change is to extract the client ID from the client assertion in the appropriate scenario when hitting the `/token` endpoint, and verify the signature.
    - Added the jwks-rsa dependency to support getting the public key from the JWKS
    - The inferno JWKS as seen at https://inferno.healthit.gov/suites/custom/smart_stu2/.well-known/jwks.json is included locally. If this is a concern we can make it fetch the live one each time. Note this is the "public view" of the JWKS which doesn't include the private keys, not the full version from https://github.com/inferno-framework/smart-app-launch-test-kit/blob/main/lib/smart_app_launch/smart_jwks.json
 - Added a new "client" `SAMPLE_ASYMMETRIC_CLIENT_ID`. This client should be used only for the confidential asymmetric method. (Note that the server doesn't currently enforce the appropriate auth mechanism per client so for now this is essentially the same as the public client ID. See FI-2606)
 - Reported support for confidential asymmetric auth at the `.well-known/smart-configuration` endpoint
 - Added a couple quick unit tests
   - The inferno private key as seen in  https://github.com/inferno-framework/smart-app-launch-test-kit/blob/main/lib/smart_app_launch/smart_jwks.json is stored locally as a PEM because there's no (easy) way to load the private key directly from JWKS and sign a JWT with it
 - Cleaned up and renamed a few minor items where the meaning wasn't apparent

## Testing guidance
The Inferno tests for this already exist in the SMART App Launch STU2 suite. See test 1.2 "SMART Standalone Launch". In the run settings, "Client Authentication Method" is at the bottom and can be toggled between Public, Confidential Symmetric, and Confidential Asymmetric. All 3 should work, note for Confidential Symmetric you will need to provide a client secret such as `SAMPLE_CONFIDENTIAL_CLIENT_SECRET`
To test failure, one easy option is to mess with the client assertion in the test kit, see https://github.com/inferno-framework/smart-app-launch-test-kit/blob/bfe00fa0a18bd29bf0ef600aa443e0ea511e1f50/lib/smart_app_launch/token_exchange_stu2_test.rb#L82 .


alisawallace (approved):
I verified that the SMART App Launch Standalone Launch tests pass when selecting Confidential Asymmetric client authentication type.  I also did a brief review of the code changes and it all looked fine to me, but I'm not very familiar with this codebase so I'll defer to @Jammjammjamm and/or @arscan for their approval in this regard.

src/main/java/org/mitre/fhir/authorization/AuthorizationController.java
```diff
@@ -255,25 +251,35 @@ public ResponseEntity<String> getToken(
...
 
-      clientSecret = splitDecodedValue.length >= 2 ? splitDecodedValue[1] : "";
+        clientSecret = splitDecodedValue.length >= 2 ? splitDecodedValue[1] : "";
+      } else {
+        // if no basic auth, client id should be supplied as request param
+        clientId = clientIdRequestParam;
+      }
+    } else if (JWT_BEARER_CLIENT_ASSERTION_TYPE.equals(clientAssertionType)) {
+      // confindential asymmetric
+      DecodedJWT decodedJwt = JWT.decode(clientAssertion);
```
Jammjammjamm, line 278: Is this just decoding without validating the JWT? Because we definitely should be validating the jwt.

dehall: Right this is not currently validating the JWT which seemed consistent with the level of validation being done elsewhere, but I will add that

Jammjammjamm (approved):

PR #140, opened 2024-03-18
FI-2492: Add Location.identifier and Location.type MS elements

dehall:
# Summary
Added `Location.identifier` and `Location.type` elements in support of US Core 7.0.0 MS requirements. This "PCP87052" Location appears to be the only one that the US Core 7 test actually finds so the fields were only added here.

## Testing guidance
Make sure to clear out the DB prior to running the reference server so the resources are reloaded, or alternatively set the `FORCE_LOAD_RESOURCES` env var to true.
> You can delete the server's data by stopping the containers with `docker-compose down` and then running `docker volume rm inferno-reference-server_fhir-pgdata` to remove the existing volume.

### DIFF
--- a/resources/uscore_bundle_patient_85.json
+++ b/resources/uscore_bundle_patient_85.json
@@ -252,6 +252,24 @@
         },
         "status": "active",
         "name": "PCP87052",
+        "identifier": [
+          {
+            "system": "urn:ietf:rfc:3986",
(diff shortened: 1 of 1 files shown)


yunwwang (approved):
Tested locally. All Location's MustSupport tests passed

PR #141, opened 2024-03-22
FI-2493: Add references to PractitionerRole resources

dehall:
# Summary
Adds a couple references to PractitionerRole resources that already exist in the system resources, so that the US Core test kit can find them. Previously there were no outgoing references to the PractitionerRole resources, only references from the PractitionerRole to a Practitioner.

## Testing guidance
The US Core 6.1.0 and 7.0.0-ballot test suites should now pass on PractitionerRole tests; previously they were skipped for "no data". Nothing should break on these or any versions, in particular the MedicationDispense tests shouldn't break due to these new additions.

Make sure to clear out the DB prior to running the reference server so the resources are reloaded, or alternatively set the `FORCE_LOAD_RESOURCES` env var to true.
> You can delete the server's data by stopping the containers with `docker-compose down` and then running `docker volume rm inferno-reference-server_fhir-pgdata` to remove the existing volume.


yunwwang (approved):
Tested locally. PractitionerRole test group passed

PR #142, opened 2024-03-28
FI-2605: SMART error responses should be json

dehall:
# Summary
This PR changes the output format of errors raised on the `/authorization` endpoint to be JSON per the OAuth spec. https://datatracker.ietf.org/doc/html/rfc6749#section-5.2
Previously these were rendered by the default spring error page, which meant they were HTML. (If the request came in with an `Accept: application/json` header it would return some JSON, but still not the expected format, and I don't think the spec requires that header)

## New behavior
There should be no changes in terms of pass/fail -- all requests that would previously succeed should still succeed with the exact same response, all requests that would previously return an error should still return an error, with the same http status and general string description of what went wrong. The main change is that errors will now always be of the form:

```json
{
  "error": "error_type",
  "error_description": "a short description here"
}
```

## Code changes
The key code change is just replacing the exceptions that are thrown and adding a new exception handler that renders them appropriately.  I created a new class `OAuth2Exception` which contains the error code, description, HTTP response code, and a header. (The header is only used for the specific instance of `invalid_client` when the client attempted to authenticate using basic auth, which per the spec is supposed to respond with a specific header.) The HTTP response code defaults to 400 Bad Request since this is what is used in the majority of cases.
There are a couple more specific exceptions already in the code, so to reduce changes I made these subclasses of the new OAuth2Exception. I could also remove these entirely and just use the new OAuth2Exception everywhere if people think that's cleaner.
In all cases the text of the error message should be the same, just the exception class changed.

Note the removal of the `if (!CLIENT_CREDENTIALS_GRANT_TYPE.equals(grantType)) {` block in the `getTokenByBackendServiceAuthorization` method is because that method is only ever called if the grant type is client_credentials, so that check would always be false.

## Testing guidance
All tests should pass/fail exactly as before. Some simple places to look at the error response are in the SMART App Launch test kit, STU2 test 3.2 "Authorization request fails when..."

Another easy option to test failure is to mess with the client assertion in the test kit, see https://github.com/inferno-framework/smart-app-launch-test-kit/blob/bfe00fa0a18bd29bf0ef600aa443e0ea511e1f50/lib/smart_app_launch/token_exchange_stu2_test.rb#L82 .

### DIFF
--- a/src/main/java/org/mitre/fhir/authorization/AuthorizationController.java
+++ b/src/main/java/org/mitre/fhir/authorization/AuthorizationController.java
@@ -39,11 +39,11 @@ import org.hl7.fhir.r4.model.Bundle.BundleEntryComponent;
 import org.hl7.fhir.r4.model.Encounter;
 import org.json.JSONObject;
 import org.mitre.fhir.HapiReferenceServerProperties;
-import org.mitre.fhir.authorization.exception.BearerTokenException;
 import org.mitre.fhir.authorization.exception.InvalidBearerTokenException;
 import org.mitre.fhir.authorization.exception.InvalidClientIdException;
 import org.mitre.fhir.authorization.exception.InvalidClientSecretException;
-import org.mitre.fhir.authorization.exception.InvalidScopesException;
+import org.mitre.fhir.authorization.exception.OAuth2Exception;
+import org.mitre.fhir.authorization.exception.OAuth2Exception.ErrorCode;
 import org.mitre.fhir.authorization.exception.OpenIdTokenGenerationException;
 import org.mitre.fhir.authorization.token.Token;
 import org.mitre.fhir.authorization.token.TokenManager;
@@ -55,13 +55,14 @@ import org.mitre.fhir.utils.exception.RsaKeyException;
 import org.springframework.http.CacheControl;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpStatus;
+import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.ExceptionHandler;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RestController;
-import org.springframework.web.server.ResponseStatusException;
 
 
 @RestController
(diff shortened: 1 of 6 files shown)


src/main/java/org/mitre/fhir/authorization/AuthorizationController.java
```diff
@@ -98,42 +99,32 @@ public String validateClientId(@PathVariable String clientId, HttpServletRequest
...
 
-    String clientId = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6InJlZ2lzdHJhdGlvbi10b2tlbiJ9"
-        + ".eyJqd2tzX3VybCI6Imh0dHA6Ly8xMC4xNS4yNTIuNzMvaW5mZXJuby8ud2VsbC1rbm93bi"
-        + "9qd2tzLmpzb24iLCJhY2Nlc3NUb2tlbnNFeHBpcmVJbiI6MTUsImlhdCI6MTU5NzQxMzE5NX"
-        + "0.q4v4Msc74kN506KTZ0q_minyapJw0gwlT6M_uiL73S4";
+    HapiReferenceServerProperties properties = new HapiReferenceServerProperties();
+    String clientId = properties.getBulkClientId();
     if (!clientId.equals(decodedJwt.getIssuer())) {
-      throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
+      throw new OAuth2Exception(ErrorCode.INVALID_GRANT,
```
Jammjammjamm, line 127: The error message here isn't always correct (see test 7.1.04 in g10):

Screenshot 2024-03-29 at 9 09 42 AM

dehall: Very strange. When the request comes in with header `Accept: application/json` then Spring wants to do its own translation of the response entity into JSON, even when I already marked the response entity as being that content type. The fix seems to be changing the response entity from a String containing JSON to a Map that spring can convert itself into JSON. 🤷

Jammjammjamm: That doesn't seem related to this issue. Here the error message says "Issuer should be CLIENT_ASSERTION_TYPE" instead of "Issuer should be CLIENT_ID".

The code that builds the bad request for this test:
```ruby
        post_request_content = AuthorizationRequestBuilder.build(encryption_method: bulk_encryption_method,
                                                                 scope: bulk_scope,
                                                                 iss: 'not_a_valid_iss',
                                                                 sub: bulk_client_id,
                                                                 aud: bulk_token_endpoint)

```

dehall: Oh wow I didn't even notice that. The overall format of the response in your screenshot was wrong too, that's the Spring content-type thing. Can you confirm you have the latest updates? I think I fixed that error message in my previous PR https://github.com/inferno-framework/inferno-reference-server/pull/139/files#diff-e200745d75092d4869394ec05e67be60c6952e36ff52cb043c22ed14851c451fL130

Jammjammjamm: The message is still wrong.

Screenshot 2024-03-29 at 2 13 52 PM

Jammjammjamm: Actually, nothing has changed.

Jammjammjamm (approved):

Issue #145, opened 2024-04-10
Typos in Sample Bundles make them invalid


the-kenny:
- `Bundle.entry[0].resource.id` must match `Bundle.entry[0].request.url` for `PUT`
- `Bundle.entry[].request.url`: Typo `rul` instead of `url` for some entries
- `Bundle.entry[].request.url`: Typo `Obervation` instead of `Observation` for some entries

dehall:
Thanks for reporting this - the bundles have a long history of manual modifications and apparently some typos made it in. We'll review and get those resolved

dehall:
Closing this issue as the mentioned items have been resolved in the latest release. Please let us know if you see anything else

PR #146, opened 2024-04-11
WIP: FI-2636: Upgrade to HAPI 7

dehall:
# Summary
This PR bumps the HAPI dependency to the latest version, 7.0.2. (see https://github.com/hapifhir/hapi-fhir/releases )

STILL IN PROGRESS - everything *should* work so this is ok for people to look at, but I want to do some significant cleanup of the trial-and-error that got me to this point

## New behavior
From the perspective of a client accessing a running instance of the server, there should be no changes to the behavior of any previously-available features. It is possible there are some new features that we get "for free" as a result of this upgrade.
From the perspective of a developer, you will need to use JDK 17+ instead of 11.

## Code changes
- Bump a lot of corresponding dependencies, remove the dependencyManagement items of pom.xml that were there because of old nested dependencies (we may need to update some items again after github does another security scan when this is merged)
- A few classes in HAPI were renamed but very little was significantly changed, so a few references had to be updated
- Borrowed a lot of code from the latest version of https://github.com/hapifhir/hapi-fhir-jpaserver-starter
- Replace `javax.servlet.*` with `jakarta.servlet.*` 
- This version of HAPI includes built-in support for Bulk Data, but there are two quirks that are non-compliant: (these should be fixed in HAPI as well but the faster fix is to workaround them here)
  -  getting status on a deleted job should return 404 (hapi returns 202 with a header saying "cancelled")
  - `application/ndjson` and `ndjson` are considered equivalent to `application/fhir+ndjson` in the spec but hapi doesn't accept these
  - All other bulk-data-specific code was removed, however I left in the unit test to make sure things still work as we expect
- `com.github.stefanbirkner:system-lambda` is no longer supported and doesn't work on JDK17 so I replaced it with a fork. This is used in one unit test to set an environment variable
- One of the typos mentioned in #145 caused issues so I will just fix all of them here. (only the one on line 17957 prevented startup)
- Many unit test classes have an `afterClass()` method that tries to delete their particular test data and shut down the server. If something fails in the delete process then the server doesn't shut down and every subsequent test will fail with `address in use`. I wrapped the shutdown command in a try/finally so one failing test won't make every later one also fail.


src/main/java/org/mitre/fhir/authorization/AuthorizationController.java
```diff
@@ -339,28 +337,33 @@ private ResponseEntity<String> validateCode(
...
       String rawCodeString = new String(Base64.getDecoder().decode(encodedCodeString));
       JSONObject codeObject = new JSONObject(rawCodeString);
-      String scopes = null;
-      String patientId = null;
-      String encounterId = null;
-      String codeChallenge = null;
-      String codeChallengeMethod = null;
-      String code = null;
 
+      String scopes = null;
```
dehall, line 341: Note for reviewers: This was changed because bumping the checkstyle version made it complain "too many lines between declaration and usage" for some of these vars.

resources/uscore_bundle_patient_355.json
```diff
@@ -4,7 +4,7 @@
     {
       "fullUrl": "urn:uuid:e91975f5-9445-c11f-cabf-c3c6dae161f2",
       "resource": {
-        "id": "e91975f5-9445-c11f-cabf-c3c6dae161f2",
+        "id": "355",
```
Jammjammjamm, line 7: Is it allowed for this id not to match the `fullUrl` entry?

dehall: I think it's ok per the spec since the fullUrl is a URN:
> Note that since [FHIR resources do not need to be served through the FHIR API](http://hl7.org/fhir/R4/references.html), the fullURL might be a URN or an absolute URL that does not end with the logical id of the resource (Resource.id). However, but if the fullUrl does look like a RESTful server URL (e.g. meets the [regex](http://hl7.org/fhir/R4/references.html#regex), then the 'id' portion of the fullUrl SHALL end with the Resource.id.  

http://hl7.org/fhir/R4/bundle-definitions.html#Bundle.entry.fullUrl

If you prefer though I can revert this and save it for a separate PR where the id and fullUrl and all corresponding references get updated. The point of this was for the id to match the request url since this is a PUT.

Jammjammjamm (approved):

PR #147, opened 2024-04-17
FI-2682 Add QuestionnaireResponse resources to Patient 85

yunwwang:
# Summary

This PR fixes JIRA FI-2682

## New behavior

Add four QuestionnaireResponse resources to Patient 85

## Testing guidance

* Reload data to reference server.
* Run US Core Test Suite 6.0.1 QuestionnaireResponse test group. (This test group is not automatically running)
* Verify all tests are passed

### DIFF
--- a/resources/uscore_bundle_patient_85.json
+++ b/resources/uscore_bundle_patient_85.json
@@ -18697,7 +18697,7 @@
             },
             {
               "linkId" : "/93041-2/93030-5",
-              "text" : "Has lack of transportation kept you from medical appointments, meetings, work, or from getting things needed for daily living",
+              "text" : "Has lack of transportation kept you from medical appointments, meetings, work, or from getting things needed for daily living?",
               "answer" : [{
                 "valueCoding" : {
                   "system" : "http://loinc.org",
--- a/resources/uscore_bundle_patient_85.json
+++ b/resources/uscore_bundle_patient_85.json
@@ -16629,6 +16629,18 @@
           },
           {
             "reference": "urn:uuid:bef6c601-bc94-4f55-9f87-7f46afcc2943"
+          },
+          {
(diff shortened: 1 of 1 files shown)


resources/uscore_bundle_patient_85.json
```diff
@@ -18427,6 +18450,477 @@
...
+                "valueCoding" : {
+                  "system" : "http://loinc.org",
+                  "code" : "LA30127-7",
+                  "display" : "Child care"
+                }
+              }]
+            },
+            {
+              "linkId" : "/93041-2/93030-5",
+              "text" : "Has lack of transportation kept you from medical appointments, meetings, work, or from getting things needed for daily living",
```
dehall, line 18700:
```suggestion
              "text" : "Has lack of transportation kept you from medical appointments, meetings, work, or from getting things needed for daily living?",
```

Everything currently passes when using the Inferno validator wrapper, but I happened to be working on migrating this test kit to the HL7 validator wrapper and there is one error `If text exists, it must match the questionnaire definition for linkId /93041-2/93030-5`. (You can also verify on the CLI validator, for example `java -jar ~/Downloads/validator_cli_6_3_4.jar -ig "hl7.fhir.us.core#6.1.0" ./resources/uscore_bundle_patient_85.json > 85_errors.txt`) 

I suspect that QR validation was added in a newer version of the core validator than we are using in the Inferno wrapper. From my testing, adding this question mark (to match the Questionnaire item at http://hl7.org/fhir/us/core/STU6.1/Questionnaire-prapare-example.html ) will make it pass in the HL7 validator as well as the Inferno validator.

yunwwang: Good catch. All resources are copied from examples in either US Core IG or SDC IG. Will fix this.

dehall (approved):

PR #150, opened 2024-04-23
FI-2679: Unit Test Cleanup

dehall:
# Summary
I noticed the unit tests weren't being checked by checkstyle and they were kind of all over the pleace. I enabled checkstyle, cleaned up all the test classes so there are no checkstyle errors, and cleaned up a couple other things I noticed.

## New behavior
There should be no change to any behavior. Originally I intended to only modify test classes, but I wound up making a couple light changes to the main source when I saw FhirContext being created all over the place. 

## Code changes
Generally the checkstyle errors that were fixed are:
 - line too long (> 100 chars)
 - imports in wrong order (should be alphabetical order, java.* doesn't go above the rest)
 - wrong indentation
 - missing javadoc, for public methods > some length
 - too many lines between variable declaration and first use
 - too many capital letters in a row in name (eg "testJSON" is not allowed, it has to be "testJson")

Other specific changes:
- Deleted `src/test/java/org/mitre/fhir/utils/HapiFhirH2Dialect.java`, this was a port of a HAPI fix that should have been removed as part of the HAPI 7 upgrade
- Deleted `src/test/java/org/mitre/fhir/utils/TestRSAUtils.java`, this was exactly identical to `src/test/java/org/mitre/fhir/authorization/TestRsaUtils.java`
- Created a single constant for the R4 FhirContext, since the guidance from the HAPI docs is that creating these is expensive. From https://hapifhir.io/hapi-fhir/apidocs/hapi-fhir-base/ca/uhn/fhir/context/FhirContext.html :
> Performance: This class is expensive to create, as it scans every resource class it needs to parse or encode to build up an internal model of those classes. For that reason, you should try to create one FhirContext instance which remains for the life of your application and reuse that instance. Note that it will not cause problems to create multiple instances (ie. resources originating from one FhirContext may be passed to parsers originating from another) but you will incur a performance penalty if a new FhirContext is created for every message you parse/encode. 

## Testing guidance
If you want to see the full list of checkstyle errors, go back to the main branch and add just the `<includeTestSourceDirectory>true</includeTestSourceDirectory>` bit to the checkstyle plugin configuration section in pom.xml, then run `mvn checkstyle:check`.


Shaumik-Ashraf (approved):
`mvn checkstyle:check` worked
`mvn test` worked
`CUSTOM_BEARER_TOKEN=FOO mvn jetty:run` worked
`docker compose up` worked

PR #154, opened 2024-05-13
FI-2735: Restore previous url matching behavior - fixes landing page

dehall:
# Summary
Fixes the landing page at /reference-server/ by restoring the previous trailing slash behavior.

## New behavior
Behavior should match previous behavior prior to the HAPI 7 upgrade in 3.0.0 -- URLs with or without a trailing slash should be handled the same.

## Code changes
Changed the `UseTrailingSlashMatch` setting to true to match what the default what prior to Spring 6.
Note: this setting is deprecated, but in the interest of a smaller change this is the easiest way to ensure things work how they used to. The alternative is to update all mappings to include both paths: with and without trailing slash. And I wasn't clear from the spec whether things like `./well-known/smart-configuration` should explicitly allow a trailing slash.

## Testing guidance
The main URL in question is http://localhost:8080/reference-server vs http://localhost:8080/reference-server/ . Everything else should continue to work as in prod today.


arscan (approved):
Tried it out locally and it behaved as advertised!

PR #155, opened 2024-06-05
FI-2808: DTR Bundle

tstrass:
# Summary

Add bundle to support DTR test kit Respiratory Assist Device questionnaire workflow. This includes all the resources necessary for a DTR SMART App to pre-populate the questionnaire.

### DIFF
--- a/resources/dtr_bundle_patient_pat015.json
+++ b/resources/dtr_bundle_patient_pat015.json
@@ -128,16 +128,10 @@
         ],
         "qualification": [
           {
-            "identifier": [
-              {
-                "system": "http://example.org/UniversityIdentifier",
-                "value": "12345"
-              }
-            ],
             "code": {
               "coding": [
                 {
-                  "system": "http://terminology.hl7.org/CodeSystem/v2-0360/2.7",
+                  "system": "http://terminology.hl7.org/CodeSystem/v2-0360",
                   "code": "MD",
                   "display": "Doctor of Medicine"
                 }
@@ -173,7 +167,7 @@
             "type": {
               "coding": [
                 {
-                  "system": "http://hl7.org/fhir/coverage-class",
+                  "system": "http://terminology.hl7.org/CodeSystem/coverage-class",
                   "code": "plan"
                 }
               ]
@@ -1207,7 +1201,7 @@
           {
             "coding": [
               {
-                "system": "http://hl7.org/fhir/condition-category",
+                "system": "http://terminology.hl7.org/CodeSystem/condition-category",
                 "code": "encounter-diagnosis",
                 "display": "Encounter Diagnosis"
               }
@@ -1259,7 +1253,7 @@
           {
             "coding": [
               {
-                "system": "http://hl7.org/fhir/condition-category",
+                "system": "http://terminology.hl7.org/CodeSystem/condition-category",
                 "code": "encounter-diagnosis",
                 "display": "Encounter Diagnosis"
               }
@@ -1311,7 +1305,7 @@
           {
             "coding": [
               {
-                "system": "http://hl7.org/fhir/condition-category",
+                "system": "http://terminology.hl7.org/CodeSystem/condition-category",
                 "code": "encounter-diagnosis",
                 "display": "Encounter Diagnosis"
               }
--- a/resources/dtr_bundle_patient_pat015.json
+++ b/resources/dtr_bundle_patient_pat015.json
@@ -1,0 +1,1332 @@
+{
+  "type": "transaction",
+  "entry": [
+    {
+      "resource": {
+        "resourceType": "Patient",
+        "id": "pat015",
+        "gender": "male",
+        "birthDate": "2015-02-23",
+        "address": [
+          {
+            "use": "home",
+            "type": "both",
+            "state": "MA",
+            "city": "Bedford",
+            "postalCode": "01730",
+            "line": [
+              "202 Burlington Road"
+            ]
+          }
+        ],
+        "name": [
+          {
+            "use": "official",
+            "family": "Oster",
+            "given": [
+              "William",
+              "Hale",
+              "Oster"
+            ]
+          }
+        ],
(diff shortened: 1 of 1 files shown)


resources/dtr_bundle_patient_pat015.json
```diff
@@ -0,0 +1,1332 @@
...
+                "display": "Encounter Diagnosis"
+              }
+            ]
+          }
+        ],
+        "code": {
+          "coding": [
+            {
+              "system": "http://hl7.org/fhir/sid/icd-10-cm",
+              "code": " G30.0",
```
dehall, line 1277:
```suggestion
              "code": "G30.0",
```

resources/dtr_bundle_patient_pat015.json
```diff
@@ -0,0 +1,1332 @@
+{
+  "type": "transaction",
+  "entry": [
+    {
```
dehall, line 4: None of the entries have a `fullUrl` and the validator is returning an error `Relative Reference appears inside Bundle whose entry is missing a fullUrl` . If this file can be loaded as normal or alternatively POSTed to the server and still work then don't worry about it

resources/dtr_bundle_patient_pat015.json
```diff
@@ -0,0 +1,1332 @@
...
+              }
+            ],
+            "text": "Laboratory"
+          }
+        ],
+        "code": {
+          "coding": [
+            {
+              "system": "http://loinc.org",
+              "code": "59417-6",
```
dehall, line 594: This is a "magic" code which means that the Observation has to conform to a base profile:  https://hl7.org/fhir/R4/oxygensat.html which derives from https://hl7.org/fhir/R4/vitalsigns.html
So this resource appears to need an `effective[x]` , category = vital-signs, and another code in the list `2708-6`

resources/dtr_bundle_patient_pat015.json
```diff
@@ -0,0 +1,1332 @@
...
+              }
+            ],
+            "text": "Vital Signs"
+          }
+        ],
+        "code": {
+          "coding": [
+            {
+              "system": "http://loinc.org",
+              "code": "59408-5",
```
dehall, line 848: Same notes about the "magic code" and requirements here

resources/dtr_bundle_patient_pat015.json
```diff
@@ -0,0 +1,1332 @@
...
+            "text": "Laboratory"
+          }
+        ],
+        "status": "final",
+        "code": {
+          "coding": [
+            {
+              "system": "http://loinc.org",
+              "code": "2703-7",
+              "display": "Oxygen (BldA) [Partial pressure]"
```
dehall, line 709:
```suggestion
              "display": "Oxygen [Partial pressure] in Arterial blood"
```
Code display doesn't match, this can be a warning but is an error by default

dehall (commented):
I ran this through the validator at https://validator.fhir.org/and there are a few issues. I'm not worried about all of them but in general I think we want the files to be as clean as possible

karlnaden (approved):
agree with Dylan's requests

tstrass:
> I ran this through the validator at https://validator.fhir.org/and there are a few issues. I'm not worried about all of them but in general I think we want the files to be as clean as possible

@dehall I fixed all the ones you pointed out. I still get some errors for URLs not resolving. Do you think these matter? I pulled these fixtures from the DRLS reference implementation and am worried that the CQL execution relies on these URLs.

image

dehall:
Hm yeah it's worth another look at those. For example the system `condition-category` ones should probably be replaced with `http://terminology.hl7.org/CodeSystem/condition-category`

tstrass:
> Hm yeah it's worth another look at those. For example the system `condition-category` ones should probably be replaced with `http://terminology.hl7.org/CodeSystem/condition-category`

Thanks for the tip on that. I fixed the remaining errors. There's still a bunch of warnings though. Do you think we care about those?

karlnaden:
should the fullUrl elements be removed? looks like they don't match the ids of the instances or the put statements and aren't referenced anywhere. 

I think the other warnings are fine to ignore at this time

resources/dtr_bundle_patient_pat015.json
tstrass: I gave them UUIDs, because that's what other bundles in the repo did, but then I realized that gives validator warnings that the UUIDs don't match the resource IDs, so I updated those https://github.com/inferno-framework/inferno-reference-server/pull/155/commits/1d15454bcbc7f48b6036c6ce5605d438ad895d9f

tstrass:
> should the fullUrl elements be removed? looks like they don't match the ids of the instances or the put statements and aren't referenced anywhere.
> 
> I think the other warnings are fine to ignore at this time

See https://github.com/inferno-framework/inferno-reference-server/pull/155#discussion_r1628361232

karlnaden:
> should the fullUrl elements be removed? looks like they don't match the ids of the instances or the put statements and aren't referenced anywhere.

I see now that there was originally an error around not having fullUrls. The updated approach still doesn't look great (you're calling something a uuid when it isn't, e.g., pat015), but I think we should move on rather than get this perfect

resources/dtr_bundle_patient_pat015.json
tstrass: Actually, it feels like I should revert https://github.com/inferno-framework/inferno-reference-server/commit/1d15454bcbc7f48b6036c6ce5605d438ad895d9f and live with the warnings and be consistent with the other bundles.

karlnaden: Seems fine to me

dehall (approved):
Looks good to me from a plain FHIR perspective and from a g10 perspective (nothing breaks or causes any new failures). I haven't confirmed this against any other test kits though so I have to defer to others on whether this addresses the specific needs of the test kit it's intended for

PR #163, opened 2024-07-03
FI-2867: New scope logic

dehall:
# Summary
This PR introduces some special new logic related to granular scopes, and the interaction between SMART v1 and SMART v2 scopes.

## New behavior
1. When the client requests a resource-level scope, the user should be offered the alternative of selecting granular scopes instead. As currently implemented, the user should not be able to select both a resource-level scope and granular scope at the same time (though this can be easily bypassed by requesting multiple scopes in various ways)
2. If any SMART v2 scopes (such as the granular scopes introduced in 1) are selected, all scopes will be converted to SMART v2. This check happens both in the UI, so that the user can see live exactly what scope they will be granting, and in the backend to be safe.

## Code changes
- Added new `/supportedScopes` endpoint which is fetched by the UI. This endpoint takes the scopes that were requested, and responds with a list of details about the scopes that are offered. The UI parses this to display the checkboxes to the user.
- I've tried to minimize duplication in the UI and backend as much as possible in terms of scope parsing and converting.
- The set of search parameters that are appended to granular scopes are stored in a YAML file

## Testing guidance
Probably the easiest way to test this is with the US Core test kit. The SMART App Launch group at the start of the test allows you to enter the set of scopes you want, then you can view the scopes that were actually granted in test 1.3.2.06 "Token exchange response body contains required information encoded in JSON" . Use the SMART v1 test as-is to try out the switching versions back & forth based on which scopes are selected.

### DIFF
--- a/src/main/resources/scope_parameters.yml
+++ b/src/main/resources/scope_parameters.yml
@@ -1,0 +1,15 @@
+---
+Condition:
+  - category=http://hl7.org/fhir/us/core/CodeSystem/condition-category|health-concern
+  - category=http://terminology.hl7.org/CodeSystem/condition-category|encounter-diagnosis
+  - category=http://terminology.hl7.org/CodeSystem/condition-category|problem-list-item
+
+Observation:
+  - category=http://hl7.org/fhir/us/core/CodeSystem/us-core-category|sdoh
+  - category=http://terminology.hl7.org/CodeSystem-observation-category|social-history
+  - category=http://terminology.hl7.org/CodeSystem/observation-category|laboratory
+  - category=http://terminology.hl7.org/CodeSystem/observation-category|survey
+  - category=http://terminology.hl7.org/CodeSystem/observation-category|vital-signs
+
+DocumentReference:
+  - category=http://hl7.org/fhir/us/core/CodeSystem/us-core-documentreference-category|clinical-note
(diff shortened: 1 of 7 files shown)


src/main/resources/scope_parameters.yml
```diff
@@ -0,0 +1,15 @@
+---
+Condition:
+  - category=http://hl7.org/fhir/us/core/CodeSystem/condition-category|health-concern
+  - category=http://terminology.hl7.org/CodeSystem/condition-category|encounter-diagnosis
+  - category=http://terminology.hl7.org/CodeSystem/condition-category|problem-list-item
+
+Observation:
+  - category=http://hl7.org/fhir/us/core/CodeSystem/us-core-category|sdoh
+  - category=http://terminology.hl7.org/CodeSystem-observation-category|social-history
```
Jammjammjamm, line 9:
```suggestion
  - category=http://terminology.hl7.org/CodeSystem/observation-category|social-history
```

Jammjammjamm (approved):
This seems to work as intended. When requesting a v1 or v2 resource-level scope and selecting the search param scopes, access is correctly limited to the selected search-param scopes.

I think @arscan should have an opportunity to comment on the scope selection UI.

dehall:
Just got a chance to discuss this with @arscan - decision is I'll add a couple lines of text that describe the new behavior. One line will only show up if the request included any v1 scopes and say all v1 scopes will be converted to v2 if a granular scope is selected. One will always show up and will say the granular scope checkboxes are enabled when you uncheck the corresponding resource-level checkbox

PR #176, opened 2024-10-16
FI-3258: Add CORS support per SMART 2.2

dehall:
# Summary
Enables CORS on the reference server per SMART 2.2: https://hl7.org/fhir/smart-app-launch/STU2.2/app-launch.html#considerations-for-cross-origin-resource-sharing-cors-support

> Servers that support purely browser-based apps SHALL enable [Cross-Origin Resource Sharing (CORS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) as follows:
> * For requests from any origin, CORS configuration permits access to the public discovery endpoints (.well-known/smart-configuration and metadata)
> * For requests from a client’s registered origin(s), CORS configuration permits access to the token endpoint and to FHIR REST API endpoints

In this case, we don't really want to implement a full client registration process, so we interpret the current Origin field as a client's "registered origin".

## New behavior
CORS headers should now appear on responses from pretty much all endpoints. I say "pretty much" because I didn't add CORS settings to all of our controllers such as the launch controllers which I don't think need them. I can easily add them if we want to do that.

As noted above, there are two categories of endpoints: the "public" discovery endpoints and "private" API endpoints. Responses from "public" endpoints will receive the header `Access-Control-Allow-Origin: *` (with the literal * wildcard). Responses from "private" endpoints will receive the header `Access-Control-Allow-Origin: <Origin from request>` where the value is the value of the Origin field in the request echoed back. 

## Code changes
There was a CORS interceptor bean previously in MitreServerConfig but it didn't actually do anything. The settings from that were copied as the basis for the new "PathBasedCorsInterceptor" class - this class allows for different CORS settings based on URL. 

Unfortunately the Interceptor is a HAPI concept so it only applies on HAPI-managed endpoints. For our controllers (specifically here, the controllers behind the `/token` and `/.well-known/*` endpoints) I add the Spring `@CrossOrigin` annotation with appropriate settings:

`@CrossOrigin(origins = {"*"}, allowCredentials = "false")` for "public" endpoints

`@CrossOrigin(originPatterns = {"*"}, exposedHeaders = {"*"}, allowCredentials = "true")` for "private" endpoints

(I really wanted to find a global approach for CORS so we didn't have to implement it per-controller and separately for HAPI, but the suggestions I found, eg https://spring.io/blog/2015/06/08/cors-support-in-spring-framework , I couldn't get to work)

## Testing guidance
CORS tests are being added to the SMART test kit in this PR: https://github.com/inferno-framework/smart-app-launch-test-kit/pull/75
You can also test the http responses directly with something like Postman (I use Bruno https://github.com/usebruno/bruno because it doesn't require an account)

## Open questions
 - Do we want an env var or any kind of config setting to "disable" CORS or have other options along those lines? In practice this should be as permissive today as previously

### DIFF
--- a/src/main/java/org/mitre/fhir/MitreJpaServer.java
+++ b/src/main/java/org/mitre/fhir/MitreJpaServer.java
@@ -23,6 +23,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
 import org.apache.commons.io.FileUtils;
@@ -37,6 +38,7 @@ import org.mitre.fhir.bulk.BulkInterceptor;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.ApplicationContext;
 import org.springframework.web.context.support.SpringBeanAutowiringSupport;
+import org.springframework.web.cors.CorsConfiguration;
 
 /**
  * MitreJpaServer configures the server.
@@ -155,6 +157,18 @@ public class MitreJpaServer extends RestfulServer {
     if (readOnly == null || Boolean.parseBoolean(readOnly)) {
       registerInterceptor(new ReadOnlyInterceptor());
     }
+
+    // IMPORTANT: interceptors only run on HAPI-managed endpoints,
+    // so for our own endpoints such as .well-known/, etc, we need to handle those separately
+    // See for example AuthorizationController and WellKnownAuthorizationEndpointController
+    LinkedHashMap<String, CorsConfiguration> corsConfig = new LinkedHashMap<>();
+    // /metadata is "public"
+    corsConfig.put("/reference-server/r4/metadata",
+        PathBasedCorsInterceptor.publicDiscoveryEndpointConfig());
+    // everything else is "private". note the double asterisk is necessary to match eg '/Patient/85'
+    corsConfig.put("/reference-server/r4/**",
+        PathBasedCorsInterceptor.privateApiEndpointConfig());
+    registerInterceptor(new PathBasedCorsInterceptor(corsConfig));
 
     // enable Bulk Export
     registerProvider(bulkDataExportProvider);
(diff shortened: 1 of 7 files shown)


src/main/java/org/mitre/fhir/MitreJpaServer.java
```diff
@@ -156,6 +158,18 @@ protected void initialize() throws ServletException {
       registerInterceptor(new ReadOnlyInterceptor());
     }
 
+    // IMPORTANT: interceptors only run on HAPI-managed endpoints,
```
Jammjammjamm, line 161: Could/Should those endpoints be managed by HAPI?

dehall: Hmm I hadn't even considered that possibility. I kind of suspect it isn't possible to do that, but HAPI does have support for loading in your own providers and interceptors so maybe there is a way to do all the things we want to do with HAPI's offerings. I'll explore that a little more

emichaud998:
Leaving a comment to note that I have tested this with the CORs tests I am adding to SMART v2.2 in the smart_app_launch_test_kit and all the tests are passing as expected

src/main/java/org/mitre/fhir/MitreJpaServer.java
dehall: After digging through the docs, I'm pretty sure it's not feasible. I thought there was a chance with Plain Providers (https://hapifhir.io/hapi-fhir/docs/server_plain/resource_providers.html), but those don't allow for arbitrary URLs, just specific actions on certain resource types. (For completeness, I think it is technically possible to use an Interceptor on the "request pre-processed" pointcut, and manually redirect requests to controllers based on request url, but then we're writing our own custom request handler, so it doesn't make sense to do that) 

On the other hand, I definitely expect there should be a way to define it globally via Spring instead of via HAPI, but the ways I tried previously didn't work at all. Nothing broke, they just didn't have any apparent effect so it was tough to debug. If you think it's worth spending a little more time I can do that too.

emichaud998 (approved):
Looks good and tested this out with the g10 test kit and everything works as expected!

PR #179, opened 2024-10-24
FI-3044: Enforce the appropriate auth mechanism for each client

dehall:
# Summary
The reference server has 3 conceptual clients - public, confidential asymmetric, and confidential symmetric. Previously the only authentication performed was checking the client secret if the ID provided was the confidential symmetric id, but it was possible to use the public or confidential asymmetric clients with any client secret, or the confidential asymmetric client without a client assertion. This PR enforces the appropriate auth mechanism for each client.  

This issue was identified on g10, which originally wasn't sending the token request correctly. The corresponding fix there was https://github.com/onc-healthit/onc-certification-g10-test-kit/pull/549 which was included in v6.0.1 (or in other words, the last version without the fix is v6.0.0) 

## New behavior
Clients can only authenticate with the appropriate mechanism.
More specifically this means that public clients may no longer provide credentials and confidential asymmetric clients must provide the appropriate client assertion. The server was already validating the assertion if it was provided, so what's new is just checking that it's there at all.

## Code changes
There's an existing function that checked the confidential symmetric client secret - added the extra logic there.

Also updated a unit test because for some reason it was putting the client id in the Authorization header for a public client, and that's not supposed to be there. (https://build.fhir.org/ig/HL7/smart-app-launch/example-app-launch-public.html#step-2-launch)

## Testing guidance
Run the g10 test suite and everything should pass (minus the expected TLS tests running locally) with the reference server preset but change the server endpoint to your local running instance. The key tests/groups to confirm are:
 - 1.4.05 (uses confidential symmetric client)
 - 9.2.05 (public client)
 - 9.12.2.05 (confidential asymmetric client)
   - this is the test that fails on the prod reference server today
   - on the other hand, the version of g10 prior to its corresponding fix (v6.0.0) will fail running against this updated server and pass prod today

These tests should also fail if you swap the client IDs around in the test config, for example, if you specify `SAMPLE_ASYMMETRIC_CLIENT_ID` (with any secret) as the Standalone Client ID in test group 1, it will pass on prod today but the token request should fail here with a 401

### DIFF
--- a/src/main/java/org/mitre/fhir/authorization/AuthorizationController.java
+++ b/src/main/java/org/mitre/fhir/authorization/AuthorizationController.java
@@ -335,7 +335,7 @@ public class AuthorizationController {
     }
 
     authorizeClientId(clientId, basicHeader != null);
-    authenticateClientIdAndClientSecret(clientId, clientSecret);
+    authenticateClient(clientId, basicHeader, clientAssertionType, clientAssertion, clientSecret);
 
     return clientId;
   }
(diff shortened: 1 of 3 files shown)


vanessuniq (approved):
looks good, g10 tests work as expected for each auth mechanism

PR #180, opened 2024-10-24
Fi 2727 inferno ci us core

Shaumik-Ashraf:
# Summary

Inferno CI on GitHub Actions with `inferno execute` CLI and US Core.

This is forked from PR #167.

## Testing guidance

# See [below](https://github.com/inferno-framework/inferno-reference-server/pull/180#issuecomment-2436508631) for updates


Shaumik-Ashraf:
# Updated Summary

This PR adds US Core v3.1.1 partial Single Patient API tests into CI. The excluded tests are Capability Statement group (because TLS tests cannot pass in CI) and Encounter, Organization, Practitioner, Provenance groups because of a bug that seems to be in the US Core Test Kit (FI-3354 will track this).

This PR is still provides a functional Inferno CI, and we can work towards adding those remaining groups and the other US Core versions incrementally if that's okay with everyone else.

This PR also replaces `wait-for-it` with `wait-for`, and addresses some of the feedback on the sister PR for g10.

PR #184, opened 2024-11-04
FI-2868: Add scopes_supported to smart-configuration

dehall:
# Summary
Adds the `scopes_supported` field to the well-known smart-configuration endpoint, per:
https://hl7.org/fhir/us/core/STU7/scopes.html#additional-us-core-requirements

## New behavior
The smart-configuration endpoint now contains a `scopes_supported` field with the set of scopes the server supports. This includes "launch/patient", "openid", "fhirUser", "offline_access", and the set of all SMART 2.x resource-level scopes and granular scopes expected per US Core 7. **Note** this does not currently include any wildcard scopes or SMART v1 style scope. These would be easy to add, but the field is not required to be exhaustive so I've left them out. Happy to add them if people prefer.

## Code changes
The list is generated programmatically in the WellKnownAuthorizationEndpointController -- all resource types are theoretically supported so rather than hardcode a giant list, we iterate over the list of resourceTypes and construct the scopes. The granular scopes were already defined in `src/main/resources/scope_parameters.yml` and the snippet that reads that file was copied from AuthorizationController

## Testing guidance
There are currently no Inferno tests that I'm aware of for this, so just hit the endpoint http://localhost:8080/reference-server/r4/.well-known/smart-configuration and confirm the field is as expected

### DIFF
--- a/src/main/java/org/mitre/fhir/wellknown/WellKnownAuthorizationEndpointController.java
+++ b/src/main/java/org/mitre/fhir/wellknown/WellKnownAuthorizationEndpointController.java
@@ -5,8 +5,10 @@ import jakarta.annotation.PostConstruct;
 import jakarta.servlet.http.HttpServletRequest;
 import java.io.IOException;
 import java.security.interfaces.RSAPublicKey;
+import java.util.ArrayList;
 import java.util.Base64;
 import java.util.List;
+import java.util.Map;
 import org.json.JSONArray;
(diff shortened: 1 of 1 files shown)


yunwwang (commented):
Tested with US Core Test Kit for v7.0.0

Test 1.3.1.03 Well-known configuration declares support for Additional US Core Required capabilities passed

yunwwang (approved):

PR #188, opened 2024-11-21
Add resources to DTR Bundle

elsaperelli:
# Summary
Adds resources to the existing `dtr_bundle_patient_pat015.json` file in order to test the DTR Light EHR Suite ([this PR](https://github.com/inferno-framework/davinci-dtr-test-kit/pull/26)). The ids for these resources are used in the modified `inferno_dtr_server_suite.json` present for the DTR Light EHR Suite.

## New behavior
- Should not change behavior in any other test suites where this is used.

## Code changes
- Adds the following resources to the file:
    - DTR QuestionnaireResponse 
    - CRD CommunicationRequest
    - CRD MedicationRequest
    - CRD NutritionOrder
    - CDex Task
    - CRD VisionPrescription
- Note: these resources were taken from their respective IG examples and modified to fit the existing pat015 patient if necessary. 

## Testing guidance
- @tstrass since you originally created this bundle, please make sure these additions don't break anything you use this for 😄 
- Run this branch locally and run the [this branch](https://github.com/inferno-framework/davinci-dtr-test-kit/pull/26) of the  `davinci-dtr-test-kit`, and run the DTR Light EHR Profiles Tests in the DTR Light EHR Test Suite using the Inferno Reference Server preset.

### DIFF
--- a/resources/dtr_bundle_patient_pat015.json
+++ b/resources/dtr_bundle_patient_pat015.json
@@ -15,20 +15,14 @@
             "state": "MA",
             "city": "Bedford",
             "postalCode": "01730",
-            "line": [
-              "202 Burlington Road"
-            ]
+            "line": ["202 Burlington Road"]
           }
         ],
         "name": [
           {
             "use": "official",
             "family": "Oster",
-            "given": [
-              "William",
-              "Hale",
-              "Oster"
-            ]
+            "given": ["William", "Hale", "Oster"]
           }
         ],
         "telecom": [
@@ -98,9 +92,7 @@
             "state": "NY",
             "city": "Buffalo",
             "postalCode": "14210",
-            "line": [
-              "840 Seneca St"
-            ]
+            "line": ["840 Seneca St"]
           }
         ],
         "telecom": [
@@ -117,13 +109,8 @@
           {
             "use": "official",
             "family": "Doe",
-            "given": [
-              "Jane",
-              "Betty"
-            ],
-            "prefix": [
-              "Dr."
-            ]
+            "given": ["Jane", "Betty"],
+            "prefix": ["Dr."]
           }
         ],
         "qualification": [
@@ -295,11 +282,13 @@
             "rank": 1
           }
         ],
-        "location" : [{
-          "location" : {
-            "display" : "observation2c"
+        "location": [
+          {
+            "location": {
+              "display": "observation2c"
+            }
           }
-        }]
+        ]
       },
       "request": {
         "method": "PUT",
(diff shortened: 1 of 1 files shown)


karlnaden (changes requested):
With local validator issues, I tried using https://validator.fhir.org/. This turned up one issue on the CDex Task.  It is timing out on the QuestionnaireResponse, but knowing what the errors were before, I'm more confident that is right.

resources/dtr_bundle_patient_pat015.json
```diff
@@ -1374,7 +1363,1506 @@
...
+            },
+            "identifier": [
+              {
+                "use": "usual",
+                "type": {
+                  "coding": [
+                    {
+                      "system": "http://hl7.org/fhir/us/davinci-hrex/CodeSystem/hrex-temp",
+                      "code": "UMB",
+                      "display": "Unique Product-independent Payer Person Identifier"
```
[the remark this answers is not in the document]
elsaperelli: image
The reason why I changed it is because I strangely get this error when its "Member Number"...should I just stick with that for now?

resources/dtr_bundle_patient_pat015.json
```diff
@@ -1374,7 +1363,1506 @@
...
+          "text": "Coded Attachment Request"
+        },
+        "for": {
+          "reference": "#patient"
+        },
+        "authoredOn": "2022-06-17T16:16:06Z",
+        "lastModified": "2022-06-17T16:16:06Z",
+        "requester": {
+          "identifier": {
+            "system": "http:/inferno.healthit.gov/reference-server/r4/cdex/payer/payer-ids",
```
dehall, line 2614: I'm not looking this closely at everything, I just happened to notice this scrolling through
```suggestion
            "system": "http://inferno.healthit.gov/reference-server/r4/cdex/payer/payer-ids",
```

resources/dtr_bundle_patient_pat015.json
```diff
@@ -1374,7 +1363,1506 @@
...
+          }
+        ],
+        "status": "draft",
+        "created": "2014-06-15",
+        "patient": {
+          "reference": "Patient/pat015"
+        },
+        "dateWritten": "2014-06-15",
+        "prescriber": {
+          "reference": "Practitioner/example"
```
dehall, line 2745: This is technically ok but I'd recommend making the references actually resolve

resources/dtr_bundle_patient_pat015.json
elsaperelli: Thank you! I will check for more of those

resources/dtr_bundle_patient_pat015.json
elsaperelli: Perhaps a dumb question, but in order for it to resolve that would require me adding a Practitioner resource to this bundle?

dehall: Yeah for this you could add a new one, or you could reference an existing one like `Practitioner/pra1234`, or if the field isn't required and removing it doesn't break any must-support type tests, you could do that. Whether you add a new one or use the existing one depends on how clinically realistic that would be vs how realistic it needs to be. (I assume just referencing the existing one where possible is fine)

karlnaden: I agree with referencing an existing one

tstrass (commented):
Tested with SMART App Suite, and confirmed this does not break anything.

resources/dtr_bundle_patient_pat015.json
elsaperelli: Added references to existing resources!

resources/dtr_bundle_patient_pat015.json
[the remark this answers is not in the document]
elsaperelli: I think @karlnaden and I decided to just remove the display altogether, if you agree with that too?

resources/dtr_bundle_patient_pat015.json
elsaperelli: The only reference I didn't change was the `allergyIntolerance` on `NutritionOrder`- is that okay as is or should I add an AllergyIntolerance resource so the reference resolves?

dehall: I wouldn't hold this PR up just for that but I do think it would be nicer to include one. There are some examples you can copy from in patient 355 https://github.com/inferno-framework/inferno-reference-server/blob/main/resources/uscore_bundle_patient_355.json

elsaperelli: Added!

karlnaden (approved):
ran these against the light DTR tests and they look good now.

PR #193, opened 2024-12-06
adding ccrr resources

rpassas:
# Summary
Adding a transaction bundle for the central cancer registry. It includes everything from patient 85 since we need to cover the US Core resources + some mCode resources.

## Code changes
Added a transaction bundle with additional resources.


360dgries (approved):
Make the relation from ID to URL but otherwise looks good up until the resources derived from 85, so should be good

PR #194, opened 2024-12-06
Fi 3499 cancer resources

rpassas:
# Summary
Adding cancer-specific resources for exchanging mCode conformant data.

## Code changes
Including a new transaction bundle with mCode resources.


dehall (changes requested):
Unfortunately this new file seems to be breaking something and I'm not quite sure what yet. Hitting the `/Patient` endpoint returns 7 patients which seems right, but the patient picker only displays 1 or 2. 

Latest on main: 
image

This branch: 
image

The first time I tried it was before the rebase and file rename, and both the new record and the carin patient showed up, now it's just the new record. Files are loaded in alphabetical order so I think this puts something in a bad state and the subsequent files don't load correctly.

I'm not sure yet but my guess is that it's related to re-using ids from the other files

PR #203, opened 2025-03-03
FI-3646: Add CARIN and combo Groups

holmesie:
# Summary
- Add a group for the CARIN patient (888)
- Add a "combo" group containing patients from CARIN, PDex, and US Core

## New behavior
Additional groups to support targeting bulk exports via Group method for specific IG data

## Code changes
resources/carin_proxy_group_patient_888.json
resources/combo_proxy_group.json

### DIFF
--- a/resources/carin_proxy_group_patient_888.json
+++ b/resources/carin_proxy_group_patient_888.json
@@ -1,0 +1,16 @@
+{
+  "resourceType": "Group",
+  "id": "carin-Group",
+  "active": true,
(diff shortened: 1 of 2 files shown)


dehall (approved):
Looks good from a plain FHIR perspective, no overlaps with existing resources, etc. One thing to be aware of is the server has a limitation where it doesn't include Location resources that are referenced by other resources in the given group. This was an issue in US Core, not sure if it matters to anyone else

PR #204, opened 2025-03-05
FI-3405: Add EOB:insurer SearchParameter from Carin4BB

dehall:
# Summary
Adds the ExplanationOfBenefit:insurer SearchParameter from the Carin4BB IG. Note this comes from the latest (STU 2.1) version of the IG but functionally it's the same as the earlier versions.

## New behavior
Enables searches using the `insurer` parameter on ExplanationOfBenefit resource - both as a direct parameter eg `/ExplanationOfBenefit?insurer=123` or as an include parameter eg `/ExplanationOfBenefit?_include=ExplanationOfBenefit:insurer`

## Code changes
Just added the search param file to resources so it's loaded at startup.

## Testing guidance
Reset the DB and run the server to load the new resources. Run the following sample requests:
```
GET {BASE_URL}/ExplanationOfBenefit?_include=ExplanationOfBenefit:insurer&patient=888

GET {BASE_URL}/ExplanationOfBenefit?insurer=c4bb-Organization
```

Running against the prod instance, these will both return errors, but with the new search param loaded locally they will work and you can inspect the results to confirm they are correct. For the first one, the returned Bundle should include instances of Organization resources (at the end), and for the second one, the insurer field on all returned resources should match the provided ID.

See also test 2.3.13 in the Carin4BB test kit - it currently fails on prod ( https://inferno.healthit.gov/suites/c4bb_v200/bSHsLmOOR8w#2.3 ) but should now pass running against a local server.

### DIFF
--- a/resources/carin_2_1_SearchParameter-explanationofbenefit-insurer.json
+++ b/resources/carin_2_1_SearchParameter-explanationofbenefit-insurer.json
@@ -1,0 +1,51 @@
+{
+  "resourceType" : "SearchParameter",
+  "id" : "explanationofbenefit-insurer",
+  "meta" : {
+    "versionId" : "1",
+    "lastUpdated" : "2020-03-31T06:41:13.000+00:00"
+  },
+  "text" : {
+    "status" : "generated",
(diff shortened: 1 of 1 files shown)


emichaud998 (approved):
Just tested this out with the CARIN client tests and reference CPCDS client server and it works as expected!

PR #205, opened 2025-03-05
Bump dependencies [dependabot]

dehall:
# Summary
Bumps a few direct and nested dependencies to address dependabot security alerts. Because of how these are defined in the pom, dependabot cannot create these PRs itself.

## Testing guidance
Run g10 or any other test kit against this and make sure nothing broke


Shaumik-Ashraf (approved):
thank you!

PR #207, opened 2025-03-13
FI-2834: Configurable clients

dehall:
# Summary
The original goal of this ticket was to limit the patient picker to only Patient/85 and Patient/355 so that users don't choose other patients in a g10 demonstration, but doing that globally would break other test kits that perform a SMART launch. Instead, we decided the best approach was to allow filtering the picker based on client ID, and allow creating customized new clients without having to hardcode them.

## New behavior
By default, using an existing client IDs means everything works exactly as it did before. 

From an end user's perspective, the only difference is a new client ID `SAMPLE_G10_CONFIDENTIAL_CLIENT_ID` which is only shown patients 85 and 355 in the patient picker. 

From a developer's perspective, new clients can now be defined by dropping a JSON file in the `./clients` folder or the existing clients can be easily configured. Note that only the new clients can be customized without a new build. (I considered adding an env var or something to disable the default clients, but the UI at https://inferno.healthit.gov/reference-server/ would be tricky to update I think. If we really want that for some reason let's make it another ticket)

## Code changes

- `src/main/resources/default_clients.json`
  - Set of default client definitions, matching what we used to have hardcoded.
  - These are not currently intended to be configurable without making a new build
- `src/main/java/org/mitre/fhir/authorization/Client.java`
  - Model and repository class for a SMART-on-FHIR client. This is not very Java-like, it's more Ruby-ActiveRecord-like, so I'd be ok refactoring the repository into a separate class.
  - The model is borrowed from Spring's OAuth2 ClientRegistration class, but filtered down to only the few things we really need.
  - The repository loads the default clients at startup so they are always present
- `src/main/java/org/mitre/fhir/authorization/AuthorizationController.java`
  - It looks like more changes than it really is
  - I migrated some logic related to JWKS/JWT verification from  `validateClient` to `authenticateClient`. `authenticateClient` now contains all the logic for checking if what was provided matches the expectations for the given client
  - Changed logic throughout that was looking for hardcoded or "hardcoded" settings (ie, HapiReferenceServerProperties), now everything from the request is compared to a Client definition
  - Added a new `filterPatientPickerBundleForClient`
- `clients/g10_confidential_client.json`
  - Our first custom client, tentatively "SAMPLE_G10_CONFIDENTIAL_CLIENT_ID"
  - It's the same as "SAMPLE_CONFIDENTIAL_CLIENT_ID" other than it defines customSettings.patientPicker = [85,355]
- `src/main/java/org/mitre/fhir/MitreJpaServer.java`
  - Load any custom clients at server startup
- `hapi.properties` and `HapiReferenceServerProperties.java`
  - removed properties related to clients other than their IDs. Consider this now "choosing" a client from the corresponding `default_clients.json` instead of "defining" a client
  - Added a new property to point to the custom clients folder, similar to the resources folder. Maybe neither of these is really necessary

One important note: if there is a bad client definition, unless it's malformed JSON this doesn't try to catch it at load time. I didn't really want to duplicate logic so as of now, certain errors will only be caught when someone actually tries to use a bad client. If we really want to validate client definitions early and fail-fast, I'll see what I can do to DRY

## Testing guidance
- Run g10 against a local instance of this server. There should be no differences in the results before and after this change. (vs prod, you will probably see TLS fails everywhere since local is HTTP only.)
  - Specific test sections to note include:
  - (1) Standalone Patient App
  - (8) Multi-Patient API
  - (9.16) Public Client Launch
  - (9.21) Asymmetric Client Launch
- Re-run the 1 Standalone Patient App section, but this time replace the client ID with `SAMPLE_G10_CONFIDENTIAL_CLIENT_ID`. This time you should see only patients 85 and 355 in the patient picker. After that everything should be the same
- Try to customize or add a new client in the /clients folder. Make sure the results are expected
- Try to break it. Some ideas:
  - Use an invalid client ID
  - Use the wrong client ID and make sure things don't pass


src/main/java/org/mitre/fhir/authorization/Client.java
```diff
@@ -0,0 +1,254 @@
...
+            Client.register(c);
+          }
+        } else {
+          Client c = gson.fromJson(jsonElement, Client.class);
+          Client.register(c);
+        }
+
+      } catch (Exception e) {
+        System.err.println("Failed to load client definitions from " + file.getName());
+        throw new RuntimeException(e);
```
vanessuniq, line 115: Instead of failing on the first error,  why not log the failure and continue loading the remaining valid clients if any?

vanessuniq (changes requested):

src/main/java/org/mitre/fhir/authorization/Client.java
dehall: So I do want to prevent the server from starting if there's a bad client file. I consider that invalid configuration, and other than failing to start there's not a lot we can do to indicate something is wrong. But you raise a good point about "on the first error" -- there could be other errors later and making someone fix each one and restart just to get to the next isn't friendly. I'll update that.

src/main/java/org/mitre/fhir/authorization/Client.java
```diff
@@ -0,0 +1,254 @@
...
+   */
+  public static void load(Path clientsPath) {
+    File pathAsFile = clientsPath.toFile();
+    File[] files;
+    if (pathAsFile.isDirectory()) {
+      files = pathAsFile.listFiles((d, name) -> name.endsWith(".json"));
+    } else if (pathAsFile.exists()) {
+      files = new File[] { pathAsFile };
+    } else {
+      return;
```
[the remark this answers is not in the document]
dehall: I made this an exception too. Originally this was silent because the tests were intentionally pointing to a non-existent location to signify "don't load anything in tests". But it's cleaner to just blank out the test config location. So now if we get to this line there's an actual value that someone expected to be real but isn't.

vanessuniq (approved):

PR #201, opened 2025-02-28
FI-3755 Use Authinfo in Inferno CI

Shaumik-Ashraf:
# Summary

**WIP**

PR #208, opened 2025-04-11
FI-3866: Remove build-time constraints on bulk data cached export

dehall:
# Summary
As of #186 this server supports loading a cached version of the Group/1a bulk data $export which was intended to be easily configurable and replaceable. I realized by storing the Binary IDs in `src/main/resources` but the actual Bundle containing the Binary resources in `./resources` that there were unintended dependencies - removing the cached export Bundle would break the export feature altogether, and changing the export would require a new build. The goal here is to make it possible to remove or replace the cached bulk data export without requiring a new build.

## New behavior
From an end-user perspective, the only difference should be improved configurability, as noted above. There should be no change to any test results pointing to this server with default settings.

## Code changes
- Removed the file `src/main/resources/cached_bulk_data_export_ids.json` and updated the script `pregenerate_bulk_data_export.rb` to no longer generate it. All the data in this file is also in the Bundle containing the Binaries, it just needs to be parsed out. Also added the group ID to a meta tag in the Bundle so it's completely self-sufficient.
- Updated `BulkInterceptor` to add a new static method `cacheGroupBulkExport` instead of hardcoding the cache being for Group 1a. This and the other changes in this file theoretically allow for multiple different groups to have a cached export.
- Added a new check in `MitreJpaServer` that looks for an env var pointing to the Bundle and parses it separately and caches it in the BulkInterceptor.

## Testing guidance
I recommend testing with the g10 test kit, test group 8 Multi-Patient API. Use the reference server preset and make sure to point both the "Bulk Data FHIR URL" at the top of settings and "Token URL" in the auth info section to point to your local instance, ie, replace "https://inferno.healthit.gov" with "http://localhost:8080". Start a new session for each test run because I saw passes where I expected to see skips when re-running so I think there's some caching happening.

Some things to consider:
- Try regenerating the bulk export by starting up the server then running `ruby pregenerate_bulk_data_export.rb` . This will take a few minutes to run
- Try running without specifying the env var - it will be slow, and the "Location" resource should be missing in the export but it should otherwise pass
- Try running with env var `CACHED_BULK_DATA_EXPORT=./resources/cached_bulk_data_export.json`


dehall:
CI is failing because of the AuthInfo change -- #201 will fix that

vanessuniq (approved):
Followed testing guidance with g10, no issues found.

PR #209, opened 2025-04-17
FI-3865: Load data from data repository

dehall:
# Summary
Rework the repository so that our default/production data are not stored here, they are instead stored in a separate repository that the reference server fetches before starting up.

## New behavior
No differences in terms of FHIR server, and no differences in terms of running just the java application. When running in Docker, new env vars SOURCE_DATA_REPO and SOURCE_DATA_BRANCH allow selecting a git repository to read from.

**Important**: there are certain implications that we need to decide if we like.

1. The fetch logic is a separate step, not part of the java application, so developers who want to work on the server itself will probably need to manually run the fetch script.
2. The server is designed with a persistent database in mind, which means that if the server restarts, the contents will not automatically be reloaded.
3. The fetch runs every time and points to a branch, so if there are new commits added to the branch those will be fetched automatically the next time the server starts. However given (2), they may not be loaded.
4. `git clone` is pretty simple - it allows for specifying a branch or tag but not a commit. If we want to allow pointing to a specific commit I'll need to use additional git commands. 
5. The fetch script replaces the contents of the `./resources` and `./clients` folders, rather than adding to them. Because of some of the above considerations, I wouldn't want the default to be add, but I'd be ok making another env var to choose add vs replace, if we want that.
6. The dockerfile doesn't contain a default value for the SOURCE_DATA_REPO env var, so the docker image by default will contain no resources and it's up to the user to add their own or set the env var. This is a breaking change -- do we want to do this?

## Code changes
No changes to Java code -- see changes to Dockerfile, and new scripts `start_reference_server.sh` and `fetch_data_repo.sh`. Even though the scripts are so simple and could be combined, fetch_data_repo is separate so that the developer who wants to work locally can run it by itself.

NOTE there are 2 things here that I left in for easier testing but will revert after:
- cert installation in the dockerfile (will comment this out)
- pointing to the branch on inferno-reference-server-data

## Testing guidance
The 2 main ways to run this are with the SOURCE_DATA_REPO env var either unset, or set to `https://github.com/inferno-framework/inferno-reference-server-data` . Until https://github.com/inferno-framework/inferno-reference-server-data/pull/1 is merged, you'll want to also specify SOURCE_DATA_BRANCH `fi-3865-ref-server-data`


Jammjammjamm (commented):
> The server is designed with a persistent database in mind, which means that if the server restarts, the contents will not automatically be reloaded.
The fetch runs every time and points to a branch, so if there are new commits added to the branch those will be fetched automatically the next time the server starts. However given (2), they may not be loaded.

This is potentially confusing, but I can live with it.

> The dockerfile doesn't contain a default value for the SOURCE_DATA_REPO env var, so the docker image by default will contain no resources and it's up to the user to add their own or set the env var. This is a breaking change -- do we want to do this?

This is a tricky one. If we default it to the data repo, it will automatically delete anything a user might have mounted to the `resources` folder, right? So we either don't load anything by default or potentially delete the user's data by default?

dehall:
> This is a tricky one. If we default it to the data repo, it will automatically delete anything a user might have mounted to the `resources` folder, right? So we either don't load anything by default or potentially delete the user's data by default?

Yeah exactly. There are other possible combinations such as load from the data repo and just overwrite what's in the folder rather than delete it, but that could introduce a lot more problems I think. I prefer the "don't load anything by default" approach because the migration notes are just "if you want to load up the usual default data, either set this env var or run this script once to download it yourself".

Maybe the fetch script should be a little smarter and error out if the local resources folder contains any jsons in it, rather than just clear it out?

Jammjammjamm:
> Maybe the fetch script should be a little smarter and error out if the local resources folder contains any jsons in it, rather than just clear it out?

Not wild about that. I think I agree that just not loading anything is the correct default, and bumping the reference server to 4.0.0 and explaining this change in the release notes should be sufficient.

Jammjammjamm (approved):

